PR #5, opened 2018-07-30
CentOS7 slave base dockerimage

rattermeyer:
Support for a centos 7 jenkins slave base.
This must be accompanied by the matching ods-configuration change.


rattermeyer:
this is already merged as part of opendevstack/ods-configuration-sample#2

Issue #18, opened 2018-08-14
Jenkins slave cannot access Nexus with self signed certificate


tjaeschke:
There is a problem if Nexus uses a self signed certificate. If for example gradle is used to build an application nexus can't be accessed. The build breaks with a certificate validation exception.

```
A problem occurred configuring root project 'prov-cd-prov-app-dev'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve org.springframework.boot:spring-boot-gradle-plugin:1.5.9.RELEASE.
     Required by:
         project :
      > Could not resolve org.springframework.boot:spring-boot-gradle-plugin:1.5.9.RELEASE.
         > Could not get resource 'https://nexus-cd.192.168.99.100.nip.io/repository/jcenter/org/springframework/boot/spring-boot-gradle-plugin/1.5.9.RELEASE/spring-boot-gradle-plugin-1.5.9.RELEASE.pom'.
            > Could not GET 'https://nexus-cd.192.168.99.100.nip.io/repository/jcenter/org/springframework/boot/spring-boot-gradle-plugin/1.5.9.RELEASE/spring-boot-gradle-plugin-1.5.9.RELEASE.pom'.
               > sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
      > Could not resolve org.springframework.boot:spring-boot-gradle-plugin:1.5.9.RELEASE.
         > Could not get resource 'https://nexus-cd.192.168.99.100.nip.io/repository/maven-public/org/springframework/boot/spring-boot-gradle-plugin/1.5.9.RELEASE/spring-boot-gradle-plugin-1.5.9.RELEASE.pom'.
            > Could not GET 'https://nexus-cd.192.168.99.100.nip.io/repository/maven-public/org/springframework/boot/spring-boot-gradle-plugin/1.5.9.RELEASE/spring-boot-gradle-plugin-1.5.9.RELEASE.pom'.
               > sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
      > Could not resolve org.springframework.boot:spring-boot-gradle-plugin:1.5.9.RELEASE.
[...]

PR #19, opened 2018-08-15
Added certificate retrieval for slave base

tjaeschke:
Enhanced the slave base image witth a retreival of an certifacte for the base application domain in case of using self signed certificates. The certificate is stored inside the image in the tmp folder, so the certificate can be used by jenkins slaves, based upon this image.


tjaeschke:
Part of the fix for #18

tjaeschke:
The slave base only stores the cert. We don't have java in all jenkins slaves, so the keytool import has to happen in the respective slave. I have modified the slave image that uses gradle as well


--

OPITZ CONSULTING Deutschland GmbH

Standort Gummersbach
Torsten Jaeschke – Senior Consultant
Kirchstraße 6, 51647 Gummersbach

Telefon: +49 2261 6001-1164
Telefax: +49 2261 6001-4100

Mobil: +49 173 9002713
torsten.jaeschke@opitz-consulting.com
www.opitz-consulting.com<http://<br/>www.opitz-consulting.com>

Geschäftsführer: Bernhard Opitz
HRB-Nr. 76555 Amtsgericht Köln

----

________________________________

PR #23, opened 2018-08-21
Update CHANGELOG.md to reflect 0.2RC

clemensutschig:


CHANGELOG.md
```diff
@@ -1,7 +1,16 @@
 # Changelog
 
-## Unreleased
-
 ## 0.1.0 (2018-07-27)
 
 Initial release.
+
+## 0.2RC *current master*
```
michaelsauter: As noted elsewhere, I'd prefer to keep this under `Unreleased`.

Since we are using SemVer, there is also no proper way to know what the next version number will be (it works pre-1.0 because you don't make BC promises but post-1.0 you can't predict if the next release requires a major version bump or not ... )

PR #39, opened 2018-10-23
- Implemented: made storage data, provisioner and backup optional

tjaeschke:
Made storage data, provisioner and backup optional, so they can be empty. This is needed by Minishift, because there are no defined values for this.


jenkins/master/plugins.txt
```diff
@@ -1,6 +1,6 @@
...
 greenballs:1.15
 credentials-binding:1.12
-ansicolor:0.5.0
-email-ext:2.57.2
-token-macro:2.1
-sonar:2.6.1
\ No newline at end of file
+email-ext:2.63
+sonar:2.6.1
+ansicolor:0.5.2
```
michaelsauter: Why do we have those changes in here? Did you test those new versions?

sonarqube/ocp-config/sonarqube.yml
```diff
@@ -436,6 +436,5 @@ parameters:
 - name: REPO_BASE
   required: true
 - name: STORAGE_CLASS_DATA
-  required: true
 - name: STORAGE_PROVISIONER
-  required: true
+- name: STORAGE_CLASS_BACKUP
```
michaelsauter: Could you enable trailing newlines in your editor? That would be great. I think we actually wanted to document that somewhere ...

tjaeschke: Hi, yes, there is an open issue about coding conventions. I have activated it in my editor now.

PR #41, opened 2018-10-29
Document the different template parameters.

### LINKED ISSUE #25: document every yaml param - descriptions en gros are missing
e.g. https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/bc.yml#L72

we should use the same description as we have in 
https://github.com/opendevstack/ods-configuration-sample/blob/master/ods-core/jenkins/ocp-config/bc.env.sample 

to have this really consistent.

rattermeyer:
parameters are documented somewhat already in the configuration
directories, but not directly in the templates.
This has been fixed now.

resolves #25

### DIFF
--- a/jenkins/ocp-config/bc.yml
+++ b/jenkins/ocp-config/bc.yml
@@ -70,7 +70,10 @@ objects:
       type: Docker
     triggers: []
 parameters:
+# repository base url. In case bitbucket is used, needs to include /scm - so it's the path in front on opendevstack/..
 - name: REPO_BASE
   required: true
+# configuration which slave dockerfile to use, either .centos7 or .rhel7
 - name: JENKINS_SLAVE_BASE_IMAGE
+# Openshift application base dns - used for grabbing the root ca and adding into the slave
 - name: APP_DNS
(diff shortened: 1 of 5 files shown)


jenkins/ocp-config/bc.yml
```diff
@@ -70,7 +70,10 @@ objects:
       type: Docker
     triggers: []
 parameters:
+# repository base url. In case bitbucket is used, needs to include /scm - so it's the path in front on opendevstack/..
```
michaelsauter: You can add them directly into the YML by using the `description` key.

michaelsauter:
FYI @renedupont

jenkins/ocp-config/bc.yml
rattermeyer: was not sure. we have both styles. but I would also prefer the description, unless we have multiple lines. So, I would propose: Include a (short) description and add further information as comments

michaelsauter: Sounds good.

rattermeyer:
@michaelsauter pushed the changed templates

Issue #38, opened 2018-10-15
Webhook Proxy


michaelsauter:
Currently, every component creates two Jenkins pipelines, `dev` and `test`, as OCP resources. In addition, every repo has two webhooks (actually four, but two of them are pointless), which trigger both pipelines for every commit. Inside the pipeline, Jenkins checks whether the pipeline is responsible for the commit (test=master,dev=*) and then continues.

This has several drawbacks:
* Both pipelines are always triggered and a check has to be made (which is slightly buggy, e.g. in case of a fast-forward merge)
* The `Jenkinsfile` is always grabbed from `master` instead of from the current branch, so changes to the `Jenkinsfile` cannot be tested inside a branch
* The `dev` pipeline is responsible for all sorts of branches, so its build status is all over the place
* Since there is only one `dev` pipeline, we cannot run builds for different branches in parallel

To improve this situation, we have build a "webhook proxy" internally. This proxy provides one endpoint accepting webhooks from BitBucket and forwards them to the corresponding Jenkins pipeline (which is determined based on the branch name). If there is no corresponding pipeline yet, it will be created on the fly. Once a branch is deleted or a pull request declined/merged, the corresponding Jenkins pipeline is deleted automatically. The webhook proxy is a Go application - a single, no-dependency binary, produced from one file using just the standard library.

This works very reliable so far (tested in two initiatives for about a month) and solves all the pain points mentioned above.

To include it in OpenDevStack, the following has to be changed:
* Currently we run on proxy in the `cd` OCP namespace. This works well, but it requires that the user inside the pod can create/delete resources in all other namespaces. This should not be the case, we would rather have a proxy running in each `foo-cd` project, and then the user only needs permissions to create/delete resources within its own project.
* The binary needs to be built from the source code. Currently, the build config is triggered locally via `oc start-build webhook-proxy --from-dir . --follow --namespace cd`, which copies a binary in the current working dir into the Docker image. While this works well, it requires Go to be installed on the local system which is not something we would want to impose on an ODS user. We either need to create a Jenkins Go slave to build it there, or we would need to build the binary within the OCP build (which is probably the easier option, but increases the size of the Docker image - currently that is based on a barebones `alpine` image).
* Remove the `dev` and `test` pipelines from the OCP templates (they are not needed, the `master` pipeline will be auto-created on the first push)
* Adapt the webhooks in BitBucket to point to the proxy.

FYI @tjaeschke @rattermeyer @clemensutschig @stitakis @gerardcl

[...]

PR #48, opened 2018-11-19
Jenkins webhook proxy

michaelsauter:
This is the first step to implement #38. Once this is approved, I will open pull requests in `ods-project-quickstarters` (so that the webhook proxy service is created alongside each Jenkins) and in `ods-provisioning-app` (so that the webhooks target the proxy).


michaelsauter:
@tjaeschke @clemensutschig Updated with a basic check for the trigger secret: ca061f8. The secret is exposed to the Go program as an environment variable. I will ensure in the follow-up PR's to have this env var filled from a secret, and also given to the provisioning app.

PR #51, opened 2018-11-21
changed references from production branch to master branch to make tailor happy

vzell:
Changed references from production branch to master branch in .yml files to make tailor happy
otherwise we get errors when following the setup documentation


clemensutschig:
This is absolutely wrong. Gas to be production.. what is the issue???

PR #60, opened 2018-12-21
Fix pipeline name

### LINKED ISSUE #59: Wrong pipeline name
When the branch is named e.g. `bugfix/FOO-529-bar-6-baz`, then the webhook proxy sets the pipeline name to `-6` instead of `-529`.

michaelsauter:
* Regexes are greedy, so we need to be more specific when looking for
  the ticket ID (previously we got the last number in the string, not
  necessarily the ticket ID).
* Safer regex for the branch name (remove all problematic characters).

Fixes #59.


michaelsauter:
Replaced by #63.

PR #68, opened 2019-01-15
Feature/add test for webhook proxy

henrjk:
Enabling unit tests of webhook-proxy with travis


michaelsauter:
@henrjk So this is basically working, right? However, which version of the Travis file should we use?

michaelsauter:
@henrjk I see both `.travis` and `.travis.yml`. Can you check?

henrjk:
@michaelsauter should be good now.

PR #90, opened 2019-03-13
Clean up unused fields

michaelsauter:
Just some cleanup, noticed a stray `metadata` field ...


michaelsauter:
@gerardcl Let's do it on a case-by-case base. I wouldn't have done it here either if the `metadata:` would not have been in there ...

PR #93, opened 2019-03-19
Centralise configuration

### LINKED ISSUE #219: Centralise configuration
At the moment, ODS has a `ods-configuration-sample` which shows which configuration params need to be present. The user is supposed to create a `ods-configuration` repo in Bitbucket, and there mirror all the files and params as in the sample repo. The structure of the sample repo is such that it tries to mirror the file structure of the repo containing the corresponding template file.

That approach is not ideal:

* Due to the structure, many params need to be repeated, making them tedious to setup/update. It would be nicer to define repeated params in one file.
* Often, PRs in one repo do not contain the whole change but need to be accompanied by a PR in the sample repo

I propose:

* Moving the sample config into `ods-core`
* Defining all shared params in one file
* Providing scripts to the user to create `ods-configuration` and to manage updates to it

michaelsauter:
Closes #219.

### DIFF
--- a/configuration-sample/update
+++ b/configuration-sample/update
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+set -eu
+
+# Get directory of this script
+scriptDir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Expected location of ods-configuration
+cd "$scriptDir/../.."
+odsConfigurationLocation="$(pwd)/ods-configuration"
+cd - > /dev/null
+
+echo "Checking if ods-configuration repository exists next to ods-core ..."
+if [ ! -d "$odsConfigurationLocation" ]; then
+  echo "> Directory $odsConfigurationLocation does not exist yet. It will be created now."
+  mkdir -p $odsConfigurationLocation
+  echo "> Created directory $odsConfigurationLocation."
+else
+  echo "> Directory $odsConfigurationLocation exists."
+fi
+
+echo ""
+
+echo "Updating .env.sample files ..."
+cp *.env.sample $odsConfigurationLocation/
+echo ">.env.sample files updated."
+
+echo ""
+
+echo "Checking for missing params in .env files ... "
+cd $odsConfigurationLocation
+elementIn() {
+  local e match="$1"
+  shift
+  for e; do [[ "$e" == "$match" ]] && return 0; done
+  return 1
+}
+anyDrift=false
+for i in *.env.sample; do
+  actualFile="${i%%.sample}"
+  if [ ! -f "$actualFile" ]; then
+    echo "> WARNING! Actual param file $actualFile does not exist yet and will be created. Please review its contents carefully."
+    cp $i $actualFile
+  else
+    sampleParams=$(cat $i | grep "[A-Z1-9_]\+=" | awk -F'=' '{print $1}')
+    actualParams=($(cat $actualFile | grep "[A-Z1-9_]\+=" | awk -F'=' '{print $1"="}'))
+    for sampleParam in $sampleParams; do
+      if ! (elementIn "${sampleParam}=" "${actualParams[@]}"); then
+        anyDrift=true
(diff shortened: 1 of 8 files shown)


michaelsauter:
@clemensutschig Do you agree that this is the right direction? I talked to @tjaeschke last week and he also agrees that this is where we should go ...

metmajer:
@michaelsauter @clemensutschig @tjaeschke is there any activity on this topic? I would like to add support for Jira configuration and am not sure if I should target the existing code-base or wait for this PR.

michaelsauter:
@metmajer Target existing code-base. I want to merge this PR eventually, but we might wait until 2.0. Thanks for asking though! :)

michaelsauter:
Addresses #219.

michaelsauter:
@tjaeschke @kiwo @henrjk @clemensutschig This is now ready to be merged from my side. For now, it only ADDS new sample files, except for a few minor necessary adjustments to the param names which I am addressing in https://github.com/opendevstack/ods-configuration-sample/pull/50 to not break master. Going forward, I am going to point the rest of ODS to those new locations and then cleanup the old sample repo.

@metmajer I wrote a helper bash script to deal with the params. Maybe this is something you could hook into your `Makefile`, e.g. as `make config`?

stefanlack:
Nice work. Is this change already documented in https://github.com/opendevstack/ods-documentation and in the _local install guide_ (https://github.com/opendevstack/opendevstack.github.io/tree/master/docs/modules/local-install)?

The work on https://github.com/opendevstack/ods-core/pull/185 is not finished yet and there are many changes regarding the local installation. Therefore, an updated installation guide is needed, if this change is releated to local installation.

michaelsauter:
@stefanlack Well this PR only adds new functionality so far, so no changes are required yet. BUT I will need to update the install guide once I am removing the old configuration sample repo. I will do that then ...

The only breaking change right now should be the renaming of the params, which I address in the sibling PR https://github.com/opendevstack/ods-configuration-sample/pull/50, so that new installations from master still work while we are transitioning.

stefanlack:
@michaelsauter ok. Thank you for your answer.

sonarqube/ocp-config/sonarqube.yml
```diff
@@ -380,46 +380,40 @@ objects:
...
     type: ClusterIP
 parameters:
-- displayName: OC project
-  name: OC_PROJECT
-  description: Name of the ocp project to deploy to. This is normally the "cd" project
-  required: true
-# this must be base64 encoded.
 - displayName: Admin Password
-  name: ADMIN_PASSWORD
+  name: SONAR_ADMIN_PASSWORD_B64
```
henrjk: I think making it explicit that these are in base64 is a good thing.

However the tailor .B64 behavior for secrets could cause confusion here. 
See https://github.com/opendevstack/tailor/issues/148#issuecomment-548367383 also.

configuration-sample/update
```diff
@@ -0,0 +1,63 @@
...
+  else
+    sampleParams=$(cat $i | grep "[A-Z1-9_]\+=" | awk -F'=' '{print $1}')
+    actualParams=($(cat $actualFile | grep "[A-Z1-9_]\+=" | awk -F'=' '{print $1"="}'))
+    for sampleParam in $sampleParams; do
+      if ! (elementIn "${sampleParam}=" "${actualParams[@]}"); then
+        anyDrift=true
+        echo "> WARNING! $sampleParam is present in $i, but not in $actualFile. Please add it."
+      fi
+    done
+  fi
```
henrjk: Should the script also highlight keys no longer needed?

sonarqube/ocp-config/sonarqube.yml
michaelsauter: Yea - right now we do not use Tailor's encryption feature for ODS anyway though.

configuration-sample/update
michaelsauter: @henrjk It could, but I did not consider that as very critical so would rather have that as a follow-up. Feel free to open an issue for it!

PR #105, opened 2019-05-10
Delete associated builds in the foreground

michaelsauter:
https://github.com/openshift/origin/issues/14120 indicates that using
Foreground is a better option than Background, and there were cases in
which the builds where not cleaned up properly so far. We really do not
want this to happen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .project
+.vscode
 *.env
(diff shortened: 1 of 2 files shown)


michaelsauter:
@clemensutschig As mentioned, one can set `PROTECTED_BRANCHES` env variable in the pod, which allows to keep more or even all pipelines around.

PR #108, opened 2019-05-16
Copy files in init.groovy.d during boot from image to volume

### LINKED ISSUE #97: Jenkins build URL missing on existing Jenkins instance - set of build status not working
@michaelsauter i gave this some thought and we should add the groovy script

michaelsauter:
This ensures that those files are always present and up-todate.
Otherwise it is very hard to ship updates for those.

This has the downside that potential changes are overwritten. If you
need to keep your changes, you need to build a custom image / run
script.

Fixes #97.


clemensutschig:
@michaelsauter we could the jenkins fix aß well now.. or?

michaelsauter:
@clemensutschig Not sure what you mean?

Issue #123, opened 2019-06-11
externalize download url of sonarqube installation binary to environment variable


stitakis:
Currently sonarqube dockerfile downloads the installation binary from a hardcoded url. The url needs to be configurable in order to enable teams to download a different sonarqube distribution from a different url.

Proposed changes/solution:
- add an environment variable to the Dockerfile
- add tailor script to sonarqube 
- extend readme to explain that the environment variable has to be defined in the build config
- add example configuration to the ods-configuration-sample

stitakis:
#125

PR #126, opened 2019-06-12
Externalize sonarqube download url

stitakis:
PR for branch 1.1.x, feature #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Scripted Nexus setup ([#42](https://github.com/opendevstack/ods-core/issues/42))
 - Webhook Proxy: Allow to protect all branches or branches with certain prefix ([#55](https://github.com/opendevstack/ods-core/issues/55))
 - Add tailor CLI to Jenkins base slave ([#62](https://github.com/opendevstack/ods-core/issues/62))
+- Add Jenkins slave `nodejs10-angular`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,29 @@
 
 ## [Unreleased]
 
+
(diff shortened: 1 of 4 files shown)


clemensutschig:
@stitakis  - this is a breaking change I would say ... so it should go only into master
@michaelsauter  what do you think?

gerardcl:
@clemensutschig I moved this task to `In Progress` since this should come with the related updates in the related `ods-configuration/-sample` as aforementioned

stitakis:
thanks... changes in sample config is missing

stitakis:
@gerardcl example added as PR: https://github.com/opendevstack/ods-configuration-sample/pull/35

stitakis:
@clemensutschig @michaelsauter yes, change is a breaking one and is already deployed in BIX bitbucket production branch and openshift. Should I create a PR for version 1.2.0?

clemensutschig:
@stitakis  -  1.1. is still untouched of this .. I would NOT add it into 1.1 except there is something currently brokeen .. let me know what you think

stitakis:
@clemensutschig current BIX production branch includes this breaking change... I could introduce a backwards compatible change in BIX production branch and create a PR for 1.2 for this change, sounds good?

michaelsauter:
A couple of notes:

1. This contains commit unrelated to what you describe, so I assume it is based off the wrong branch (it contains merge commits that look like you based your branch on `production`, not `master`).
2. I think we could make this less breaking by adding the old URL as a default into the YML.
3. This is a new feature, and therefore should only go into `master`, not `1.1.x`. The `x` indicates we only want to add patches ...

stitakis:
@michaelsauter as discussed, we don´t need to introduce this change to 1.1.x... I´ll delete the PR

Issue #131, opened 2019-06-26
Automated nexus resource setup fails with Nexus latest image


tjaeschke:
There is a change in the latest Nexus image (3.17.0), released 2 days ago.
There is no longer a static default admin password, but a generated password instead, located at 

/nexus-data/admin.password

It can be retrieved on the console via 

oc rsh -n cd $(oc get pods -n cd | cut -d" " -f1 | grep "nexus") cat /nexus-data/admin.password

The scripts have to be modified, so the generated password will be used

rattermeyer:
Think we should fix the versions of tools we are using and not relying on "latest" or a varying tag.
Latest is annoying if you want to support software and a stable documentation.

tjaeschke:
We will open a new issue for a fixed version

stefanlack:
see #156

PR #132, opened 2019-06-28
#131 Bugfix Automated nexus resource setup fails with Nexus latest image

stefanlack:


tjaeschke:
We should fix the version in an additional step, but get the password with the fix provided by Stefan.
@stefanlack @clemensutschig Thoughts?

stefanlack:
closes #131

PR #134, opened 2019-06-28
Use Go's template package to render BuildConfig JSON file

### LINKED ISSUE #133: Webhook Proxy: Use Go template to render pipeline config
Right now the webhook proxy uses string replacement - it should use Go templates instead for more flexibility. This will be helpful for e.g. #61.

michaelsauter:
Instead of doing string replacements, use the templating packaging.

Also, this removes the global `server` variable to make testing easier,
and updates Go to 1.12.

Closes #133.


jenkins/webhook-proxy/pipeline.template.json
```diff
@@ -2,24 +2,24 @@
...
         "source": {
             "type": "Git",
             "git": {
-                "uri": "GIT_URI",
-                "ref": "BRANCH"
+                "uri": "{{.GitURI}}",
+                "ref": "{{.Branch}}"
             },
             "sourceSecret": {
                 "name": "cd-user-with-password"
```
michaelsauter: @metmajer @clemensutschig This is the template used to render the `BuildConfig` which represents the Jenkins pipeline. So any request to the webhook proxy which wants to create/start a pipeline needs to supply enough information to fill those slots. Further, we probably need to send some build args to Jenkins as well, right?

michaelsauter:
@henrjk Thanks for reviewing and the pointers to the tools. We should keep them in mind when we look at point 6 (templating for quickstarter generation) in https://github.com/opendevstack/ods-project-quickstarters/issues/254.

Issue #61, opened 2018-12-21
Trigger multiple pipelines for one commit


michaelsauter:
It is possible that one repo contains more than one `Jenkinsfile`. Now that we have a webhook proxy, it is relatively trivial to allow users to specify the locaction of each `Jenkinsfile`, and create one pipeline for each. This would help to deal e.g. with [multi module projects](https://spring.io/guides/gs/multi-module/).

The implementation would roughly look like this:

Allow to add a query param to the webhook, e.g. `?dirs=foo,bar`. In that case, the webhook proxy would create two pipelines, one pointing to `foo/Jenkinsfile` and the other pointing to `bar/Jenkinsfile`. Then it would trigger both pipelines.

Futher, each pipeline could check whether it really needs to run. It would do that by comparing if something in the folder changed compared to the last build of that folder. First, it would get the latest built commit sha via `oc -n foo-dev get -o template bc/be-main --template={{.spec.output.to.name}}`, and then it would check for changes via `git diff --exit-code <sha1>..<sha2> <folder>`.

@tjaeschke @rattermeyer @stitakis @clemensutschig Thoughts on this?

rattermeyer:
Hmm. Not conviced yet. Wouldn't be another option to create a "Master" Jenkinsfile in a multi-module build and call the other Jenkinsfile, like proposed here: https://stackoverflow.com/questions/44766324/invoke-a-remote-jenkins-file-from-another-jenkinsfile.
You would not get a seperate pipeline for each module, but if you have a multi-module project, I would expect that the different modules are released and deployed together. In this case, I would like to have a single pipeline for the project and only optimize build times as you have described.

michaelsauter:
@rattermeyer Nice find, did not know this is possible.

I am not convinced however that one would want to deploy the modules together. There are many who favour a monorepo over many repos, but that does not mean that you build a monolith (everything deployed at the same time) necessarily. Therfore I think it could make sense to have a different pipeline.

That said, I I'll think a bit more about your proposal. Also, we can wait until we have a few use cases and then decide on the best solution for this.

rattermeyer:
Monorepo <> Multi-module project. If we talk about a monorepo, then your idea makes sense. But a monorepo should /can contain independent projects. But a multi-module build tends to have the same version for all components.

michaelsauter:
As this came up in a slightly different discussion again, the solution has evolved slightly (@sino92 thanks for the hint!):

[...]

PR #135, opened 2019-07-01
Feature: Allow to customize jenkinsfilePath option of BuildConfig

michaelsauter:
Partially addresses #61. What is left now is that one can specify multiple webhooks with different names. How that would work best is not entirely clear to me yet, therefore I leave it out here.

FYI @sino92


michaelsauter:
Merging this now. Thinking about the follow-up, I've made up my mind to add another GET param called `component`, which overrides any component delivered via the payload. That way, one can trigger multiple pipelines for the same commit.

PR #145, opened 2019-07-10
Implement endpoint for custom payload

### LINKED ISSUE #80: Jenkins webhook proxy should create pipeline based on HTTP/ params
the webhook proxy should take HTTP (get?paramA=B&paramB=C) params on request and transform them into BC pipeline ENV params.

michaelsauter:
Closes #80.

1. Adds a new endpoint, `/custom`. Example payload is at https://github.com/BIX-Digital/ods-core/blob/be2795d3360a819fe20ef8c8a7e39c01cd974ecf/jenkins/webhook-proxy/test/fixtures/custom-payload.json.

2. Adds a `GET` param called `component` which overrides the component part of the pipeline, which is normally derived from the repository in the payload.

3. Responds to requests to paths other than `/` or `/custom` with 404.

FYI @ungerts @metmajer


metmajer:
@michaelsauter in the spirit of REST, can we please have this renamed to a resource instead of `custom`? The closest that comes to my mind would be `build`. Thoughts?

michaelsauter:
@metmajer So, `POST /builds` then? I'd be fine with that ...

metmajer:
@michaelsauter @clemensutschig the intention behind this feature was to trigger a specific release pipeline from within Jira. Ideally, Jira, or someone who wants to craft a release of an entire application  should need to knowledge about a `component` or a `branch`.

This is why I had originally requested to be able to create a *release* as opposed to a *build*. In the near future we may want to build upon this build functionality to create a release with specific assumptions. This would be facilitated if we could assume `component` and `branch` to be optional, based on some conventions.

metmajer:
@michaelsauter `builds` would be correct if we were to create multiple builds with a single POST request. Otherwise it should only be `build`. Thanks!

michaelsauter:
@metmajer Not knowing the `component` might work, but how do you envision not knowing `branch`? A pipeline runs for a repo, so it must run for a certain branch. We can default it to something, but some branch must be given ... otherwise there is no way to get to the `Jenkinsfile` ... or am I not getting the idea at all?

Regarding `build` vs `builds` - AFAIK this depends on style, not whether you create one or more of the resource. There are lots of APIs that use plurals for the "collection", and `POST`ing to the collection simply adds one resource to the collection ...

metmajer:
@michaelsauter with regards to `build` vs. `builds` there is indeed no right or wrong. I've made it a habit to use the singular if operations such as "creating" or "listing" on a collection of resources don't make sense. Never mind. That may be debatable, but not important so please proceed to use with whatever matches your existing model.

With regards to your other question on branches, let me circle back to our discussion on a project's branching strategy (branch model) we started over at https://github.com/opendevstack/ods-project-quickstarters/issues/254#issuecomment-506829194. Here's a quick example: let's assume that developers check in their code on `master`. Before a release, code has to be merged onto a `release/v$version` branch. After a release, a tag `$version` shall be created on `master`.

Neither of this should be done unorchestrated or manually, since a merge operation from one git branch to another will trigger a webhook and associated pipeline run for an ODS component. We are building the MRO pipeline and associated shared library to make sure that such actions automatically occur in the right order.

Here's my suggestion: assuming that my yet limited understanding about ODS behavior is correct, I'd like to see the MRO pipeline handle the branching and tagging operations as laid out in the following example. The configuration entries are meant to go into a project-specific `metadata.yml` that accompanies the MRO pipeline:

```
branches:
  - name: master
    when: $env == "dev"
  - from: master
    to: release/v$VERSION
    when: $env == "qa"
  - from: release/v$VERSION
    to: tag:$VERSION
    when: $env == "prod"
```

Obviously, this is only a draft and especially the last rule doesn't quite hit the nail yet. We'd have to discuss how this fits into the concepts around `branchToEnvironmentMappings` and where we would need to adapt. Because of its orchestration capabilities, the MRO pipeline should be regarded a core component of any project and moving project-specific settings into `metadata.yml` may thus be justified.

Note that we won't use this endpoint to run ODS component pipelines in a release context, as it would create multiple uncorrelated pipeline runs in Jenkins. Instead, we'll load each component's Jenkinsfile into a Jenkins Job and have it executed in the context of the MRO pipeline.

michaelsauter:
@metmajer Changed it to `/build`, I don't really mind the name ...

Regarding the `branch` param, I don't understand how what you are describing applies at all here. Remember that the webhook proxy creates a new pipeline when the pipeline which is triggered does not exist yet. When it creates a new pipeline, it does so by filling and applying https://github.com/opendevstack/ods-core/blob/bc39279117303791cb07aa176f5242643e34c05d/jenkins/webhook-proxy/pipeline.json.tmpl, which needs a branch value.

clemensutschig:
@michaelsauter  - we should default to master, one can overwrite that anyway as part of a rest call from a CLI or alike ... (to say pick up another branch). 

Regarding @metmajer (s) comment on component - as one can pass env params anyway - the pipeline can do whatever it likes behind the scenes...

metmajer:
@michaelsauter @clemensutschig I'll provide more additional context for clarification today.

michaelsauter:
I'd rather keep the `branch` input required - clients can easily default to `master`. I want the clients to understand that they are referencing a specific `Jenkinsfile`. 

`component` is not really negotiable I think as BitBucket gives us no control over the body payload, so to support multiple components per repo, we need to add it to the query params, even if that is not pretty.

I would say this PR is ready to merge - unless you see a blocker?

metmajer:
@michaelsauter Right, let's move on. We can discuss alternatives later on. Thanks!

PR #155, opened 2019-07-19
Validate request

### LINKED ISSUE #154: Webhook Proxy Bad Request handling
Right now, it always returns 200. We should not return 200 when:

- JSON cannot be parsed
- Required arguments are not given
- Pipeline should have been created but could not for whatever reason (non 2xx response code from OpenShift).

michaelsauter:
Closes #154.


jenkins/webhook-proxy/main.go
```diff
@@ -307,6 +316,11 @@ func (s *Server) HandleRoot() http.HandlerFunc {
 
 		log.Println(requestID, event)
 
+		if !event.IsValid() {
```
henrjk: Perhaps isValid could return a short error description which is returned to the caller?

michaelsauter: You are right, we could improve this. Overall, I wasn't quite sure what to check for. There are also quite a few packages that deal with validation, which might provide additional value, but I wasn't quite ready yet to make the jump from "no dependency" to "one dependency". I'll leave it like this for now, and we can improve this further, but I think that also entails some research what each individual param's values are ...

PR #168, opened 2019-08-07
Feature/bix 573 add snyk support to ods

stitakis:


rattermeyer:
Would like to open a discussion on this one. 
Snyk is a commercial tool. How would we go forward with including this into the open source version?
Other companies might come to other conclusions, what tool to use. Would very much appreciate an option to include something like https://dependencytrack.org/ into the platform, and show how to include a different (commerical) tool into the shared library / image / ...
Probably you have discussed this already, how to handle this in the open source project?

rattermeyer:
I am not blocking this, but we should really discuss this in the context of the open source project as well.

stitakis:
Hi @rattermeyer sure, no problem... thanks for raising this... Snyk is actually free to use as described here: https://snyk.io/plans/   The free plan provides less features and limited number of scans for private repositories but unlimited for open source projects. On the other side the commercial option provides more functionality and integration options.

rattermeyer:
Hi, saw this as well. My understanding was unlimited for open source. The 200 tests means: just a trial oder how do I understand this? After running 200 tests, it stops? If it was always available for a limited number of projects, that would make it an option.
I was already thinking about the fact, that it has a free plan for open source projects. But concluded that opendevstack itself will most likely not be used for open source projects itself.

michaelsauter:
Commit history is better, but still needs to be cleaned up - I think we should refrain from merging master into the branches on local BitBucket instances ... sorry to bother you with this ...

stitakis:
closing this in favour to #172, it has a better commit history

PR #180, opened 2019-09-06
Improve documentation of webhook proxy

### LINKED ISSUE #174: webhook proxy /build does not work in case repository has enabled bb webhook proxy
the `/build` endpoint today - creates a pipeline in the same name format as a request coming from the "normal" bitbucket webhook.

As there is always a webhook on a repo - `/build` will never create a new pipeline as it's already there (after a push of a new branch into bitbucket, that triggers the creation of a pipeline) - so env params will never make it... (and hence you can never "really" build)

proposal is to postfix the pipeline name with `builder` in case of `/build` being called.

@michaelsauter  - cool?

michaelsauter:
Closes #174.

@rattermeyer I don't really understand the new documentation system yet. What do I need to do to get the documentation on opendevstack.org updated? And do I need to write this in adoc? I'm a bit confused to be honest. Is there documentation about how to document? :D

### DIFF
--- a/jenkins/webhook-proxy/README.md
+++ b/jenkins/webhook-proxy/README.md
@@ -1,34 +1,65 @@
 # Jenkins Webhook Proxy
 
-Provides one endpoint accepting webhooks from BitBucket and forwards them to the
-corresponding Jenkins pipeline (which is determined based on the branch name).
-If there is no corresponding pipeline yet, it will be created on the fly. Once a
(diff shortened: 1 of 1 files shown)


rattermeyer:
You are right. We need to document it. needs push to ods-documentation. need to set up a cron job on travis.

stefanlack:
Documentation should be added at this location https://www.opendevstack.org/ods-documentation/common/latest/documentation.html

Issue #210, opened 2019-10-25
Remove image trigger from Jenkins master instances


michaelsauter:
**Is your feature request related to a problem? Please describe.**
With lots of Jenkins master instances in one cluster, having an image trigger leads to lots of re-deployments at the same time. This can have an impact on cluster performance.

**Describe the solution you'd like**
It might be better to remove the trigger and provide tooling instead to rollout new versions in batches instead of all-at-once.

**Describe alternatives you've considered**
None - potentially there is something in OCP that can stagger the image trigger?

**Additional context**
Also, not tracking the `latest` tag will help ...

clemensutschig:
done - see 
https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/deploy/jenkins-master.yml#L250

PR #212, opened 2019-10-25
Tag images with ODS version

### LINKED ISSUE #211: Tag Jenkins images with ODS version
Right now, Jenkins master and slave images are only tagged with `latest`, and all `Jenkinsfile` references point to `latest`. I think this is fragile, and propose to use a similar approach as we do for the Jenkins shared library: tag with the ODS version.

To be precise:
For a given ODS branch, e.g. `2.0.x`, I would set all `BuildConfig` items to emit tags `2.0.x`, and all quickstarter `Jenkinsfile`s point to that tag, e.g. `image: "${dockerRegistry}/cd/jenkins-slave-maven:2.0.x"`. Also, the Jenkins master should point to that image.

That would allow us to make breaking changes if needed. Also, it allows users to pin a version if a change that ODS deems "non-breaking" actually turns out to be breaking for that user.

If OpenShift supports building multiple tags, I would support to continue tagging with `latest` additionally, to offer that for users who want to be bleeding edge.

Finally, if we are concerned regarding upgrading, we could offer at least "major version tags", e.g. sth like `2.x`.

@rattermeyer @tjaeschke @clemensutschig @stitakis Please comment if that is a direction we want to go in. I propose to introduce it together with ODS 2.0.

michaelsauter:
Instead of advancing `latest`, we will now tag with the ODS version
to provide a more stable experience.

The implication of this is that with an ODS update, not all users
will automatically get the new version. They need to opt-in.

Please also see the related issue #210.

Closes #211.

@rattermeyer @clemensutschig @metmajer As this is a bigger conceptual change, I will wait with merging until you all approve this PR.

Once approved, I will create a follow-up PR in ods-project-quickstarters that makes all slave images use the same tag, and all `Jenkinsfile`s point to the ODS version tag.

### DIFF
--- a/jenkins/ocp-config/bc.yml
+++ b/jenkins/ocp-config/bc.yml
@@ -40,7 +40,7 @@ objects:
     output:
       to:
         kind: ImageStreamTag
-        name: 'jenkins-master:latest'
+        name: 'jenkins-master:2'
 - kind: BuildConfig
   apiVersion: v1
   metadata:
@@ -52,7 +52,7 @@ objects:
     output:
       to:
         kind: ImageStreamTag
-        name: 'jenkins-slave-base:latest'
+        name: 'jenkins-slave-base:2'
     postCommit: {}
     resources: {}
     runPolicy: Serial
@@ -89,7 +89,7 @@ objects:
     output:
       to:
         kind: ImageStreamTag
-        name: 'jenkins-webhook-proxy:latest'
+        name: 'jenkins-webhook-proxy:2'
     postCommit: {}
     resources: {}
     runPolicy: Serial


rattermeyer:
LGTM for what we intend to do. Not sure if we discuss this on the PR.

What I could also imagine is having multiple tags: 2, 2.0 and later 2.1 etc. on the 2.x branch if we introduce patches.

This allows users to either pin to a specific patch version or the latest "2" version.
Although we should be completely backward compatible, but there is still a small risk here, as [pointed out](https://github.com/opendevstack/ods-core/issues/211#issuecomment-534993602).

We could refer in the ODS code (and quickstarters) always to the ODS version, but document how to pin to a specific version and document the trade-offs of each decision.

michaelsauter:
@rattermeyer Unfortunately, the building of the tags is actually the responsibility of the ODS users. We as maintainers do not offer any images, so we cannot provide e.g. 2.0, 2.1 ... so users would need to tag a specific SHA with e.g. 2.0, and then can have their internal code point to it. I would argue that is completely out of our hands ...

michaelsauter:
@clemensutschig @metmajer ping

PR #223, opened 2019-11-01
airflow HTTPS no verify if env var PYTHONHTTPSVERIFY is 0

### LINKED ISSUE #222: Airflow quickstarter TLS verification can fail
**Describe the bug**
Depending of the type of OpenShift cluster used, Airflow webserver fails due to not being able to know its own domain certificates (those are not provided in the base images).

**To Reproduce**
1. Provision Airflow cluster quickstarter
2. Once all deployments are ready, go to your Airflow webserver route
3. OpenShift OAuth callback fails due to not knowing the CA (ssl error verifcation)

**Expected behavior**
If possible, required CA should be known, if not possible but trusted then one should be able to disable verification.

 - OpenShift: OCP
 - OpenDevStack 1.2.0

**Additional context**
Solution can be that in the OpenShift OAuth service implementation for Airflow we can implement, via environment variable, to disable or not the certificate verification.

gerardcl:
fixes #222


sklingel:
Who and what will set PYTHONHTTPSVERIFY? Manual environment variable set in OC or needs the template also be updated? @gerardcl
Or a quick word in the README?

gerardcl:
Hi @sklingel !
This should never come within template, the user decides to use it via OC env var.
Did not thought on adding it into the README since it is a known env var for such purpose when workin with python. But, we can add this for sure.
Can you do so? or shall I?

gerardcl:
@sklingel note added, let me know if it's ok!

gerardcl:
we are going to tackle a more airflow-ish approach in the ods-quickstarters new version

PR #231, opened 2019-11-11
Allow to configure external projects

### LINKED ISSUE #229: Allow to specify project in build endpoint of webhook proxy
The webhook proxy allows (with a minor exception for the prov app) only to manage/proxy within it's own project. This is good for security, but when Rundeck is gone, we need to access (at least read from) the `opendevstack` namespace as well.

PR #227 added a hack to master to allow building any URI. That should be reverted, and instead we should allow to build only for the current namespace **and** whatever is configured via a new env var `ALLOWED_EXTERNAL_PROJECTS`, which, if not set, defaults to `opendevstack`. The `build` endpoint payload then has to learn a new field, `project`, which allows to specify a project where the repository is located.

@kiwo I'll implement this in the next few days, you'll need to update your PR to the provisioning app then.

michaelsauter:
Closes #229.


docs/modules/jenkins/pages/webhook-proxy.adoc
```diff
@@ -47,13 +86,11 @@ The following environment variables are read by the proxy:
 | The base URL of the repository (e.g. your BitBucket host). This variable is set by the template and usually does not need to be modified.
 
 | TRIGGER_SECRET
-| The secret which protects the pipeline to be executed from outside. This variable is set by the template and usually does not need to be modified.
-|===
+| The secret which protects the pipeline to be executed from outside. This variable is set by the template and usually does not need to be modified. |
```
henrjk: When I go to View file this paragraph is not rendered as expected.

docs/modules/jenkins/pages/webhook-proxy.adoc
```diff
@@ -47,13 +86,11 @@ The following environment variables are read by the proxy:
...
 
-Moreover, one can pass the following query parameters to the proxy:
-| Variable | Description |
-| -- | -- |
-| jenkinsfile_path | The path to the `Jenkinsfile`. By default, the `Jenkinsfile` is assumed to be in the root of the repository, therefore this value defaults to simply `Jenkinsfile`. |
+| ALLOWED_EXTERNAL_PROJECTS
+| Comma-separated list of external projects which the Webhook Proxy can deal with. By default, this is just the `opendevstack` project.
```
henrjk: When I go to View file this paragraph is not rendered as expected.

michaelsauter: Thanks! The problem was the trailing `|` on the previous line ... fixed in 82d2ea92e.

michaelsauter:
@henrjk Thanks for looking into the code coverage. I extracted an issue from that: #232.

Issue #238, opened 2019-11-15
local installation failes due to vagrant error


stefanlack:
**Describe the bug**
command `vagrant up` fails with errors, since vagrant cannot install GuestAdditions.

**To Reproduce**
1. Follow installation guide on https://www.opendevstack.org/ods-documentation/common/latest/local-install/installation.html

**Expected behavior**
Vagrant installes the correct guest additions

**Screenshots**
First attempt:
![image](https://user-images.githubusercontent.com/2901438/68940115-a0acd880-07a2-11ea-8a80-7720cc4f037b.png)

Second attempt of `vagrant up`
![image](https://user-images.githubusercontent.com/2901438/68940137-aefaf480-07a2-11ea-9a61-d4ec5cc58651.png)

**Affected version (please complete the following information):**
 - vagrant version 2.2.6
- virtualbox version 6.0.14

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**

stefanlack:
closed via https://github.com/opendevstack/ods-core/pull/239

PR #239, opened 2019-11-15
bugfix #238 local installation failes due to vagrant error

stefanlack:
Use shell script for update guest addition / add vagrant cachier plugin config


stefanlack:
Thank you.

Issue #228, opened 2019-11-08
Jenkins in prov-cd cannot give image-puller permissions


michaelsauter:
**Describe the bug**
When creating projects via Jenkins, the execution fails due to:
```
oc policy add-role-to-user system:image-puller system:serviceaccount:mstest-cd:jenkins -n cd
Error from server (Forbidden): rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:prov-cd:jenkins" cannot list rolebindings.rbac.authorization.k8s.io in the namespace "cd": RBAC: clusterrole.rbac.authorization.k8s.io "self-provisoner" not found
```

**To Reproduce**
I created and triggered a new pipeline, which should create the jobs, on a branch with quite a bit of cleanup, but I think the error is unrelated to that. The invocation was basically this:
```
curl "https://webhook-proxy-prov-cd.example.com/build?trigger_secret=SECRET&jenkinsfile_path=create-projects/Jenkinsfile&component=ods-corejob-create-projects-mstest" -v \
-H 'Content-Type: application/json' \
-d '{"branch":"task/cleanup","repository":"https://cd_user@bitbucket.example.com/scm/opendevstack/ods-core.git","env":[{"name":"PROJECT_ID","value":"mstest"},{"name":"PROJECT_ADMIN","value":"mstest@example.com"},{"name":"PIPELINE_TRIGGER_SECRET","value":"SECRET"}]}'
```

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack: master only

@kiwo Did you run across this? Did you have to change anything regarding the permissions?

EDIT: Here's the branch I was trying this on: https://github.com/BIX-Digital/ods-core/tree/task/cleanup

clemensutschig:
@michaelsauter - this used needs to be able to grant provision rights to others .. (and we need to ensure that we tighten this down as much as possible)

michaelsauter:
I have added the following rolebindings:
```
$ oc adm policy add-cluster-role-to-group system:image-puller system:authenticated -n cd
$ oc adm policy add-role-to-group view system:authenticated -n cd
```

Works like a charm 🎉 

I'll need to add this to the installation instructions.

FYI @kiwo

PR #241, opened 2019-11-15
Cleanup new Jenkins jobs to create project

michaelsauter:
Fixes a few things and makes a few things prettier.

Todo:

- [x] I had to add `https://` in front of the BitBucket host to make it work in our cluster, this needs to be better so that it also works for local installations
- [x] The `Tailorfile` in `ocp-templates/ocp-config/cd-jenkins-master` still points to the old config, this needs to be updated and the pipeline trigger secret has to be dealt with (as it is now per-project)

FYI @kiwo


create-projects/Jenkinsfile
```diff
@@ -1,55 +1,62 @@
...
+        branches: [[name: '*/master']],
+        doGenerateSubmoduleConfigurations: false,
+        extensions: [[
+          $class: 'RelativeTargetDirectory',
+          relativeTargetDir: 'ods-config'
+        ]],
+        submoduleCfg: [],
+        userRemoteConfigs: [[
+          credentialsId: 'prov-cd-cd-user-with-password',
+          url: "https://${bitbucketHost}/scm/opendevstack/ods-configuration.git"
```
michaelsauter: @kiwo This is the variable I was talking about ...

michaelsauter: Could you please confirm that your `bitbucketHost`(which simply is `BITBUCKET_HOST` of the `DeploymentConfig`) has no scheme? If that is the case, I assume that Jenkins defaults to `http://` if no scheme has been given.

michaelsauter: @kiwo Also, is your `BITBUCKET_HOST` simply `192.168.56.31:7990` as in the sample config?

michaelsauter:
@kiwo I implemented the same solution for HTTPS than you used in ods-quickstarters. This is ready to be merged from my side. I verified that it works in our cluster.

michaelsauter:
This PR assumes you have done https://github.com/opendevstack/ods-core/issues/228 in your cluster.

create-projects/create-cd-jenkins.sh
```diff
@@ -73,9 +73,10 @@ tailor_update_in_dir "${OCP_CONFIG}/cd-jenkins-master" \
     "--param=PROJECT=${PROJECT_ID}" \
     --selector "template=cd-jenkins-master-template"
 
+pipelineTriggerSecretBase64=$(echo -n $PIPELINE_TRIGGER_SECRET | base64)
```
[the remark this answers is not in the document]
michaelsauter: Oh, I was unaware of that. Fixed it in 5b5ff17 so that we now assume the param to be base64 encoded already.

PR #297, opened 2019-12-02
delete-projects job 

kiwo:
fixes opendevstack/ods-provisioning-app#294


stefanlack:
Tested change localy, when called from provision app. It works and delete the 3 projects -dev, -test and -cd as expected.

PR #304, opened 2019-12-03
Set NEXUS_URL and JIRA_URL on Jenkins DC

michaelsauter:
See https://github.com/opendevstack/ods-quickstarters/pull/34.


michaelsauter:
@gerardcl This removes the sample config file with the r-shiny stuff - that is fine, right?

PR #306, opened 2019-12-03
added missing delete-components job

kiwo:
"fixes": opendevstack/ods-provisioning-app#296

here is the missing delete-components job with the scripts as mentioned in 296 included. i dont think this will work but should not be worse than before


delete-components/delete-component.sh
```diff
@@ -0,0 +1,108 @@
...
+if [ -z ${COMPONENT+x} ]; then
+    echo "COMPONENT is unset, but required";
+    exit 1;
+else echo "COMPONENT=${COMPONENT}"; fi
+
+if [ -z ${ROUTE_NAME+x} ]; then
+    echo "ROUTE_NAME is unset, using default ${COMPONENT}";
+else
+    echo "ROUTE_NAME=${ROUTE_NAME}";
+fi
```
michaelsauter: It looks like you read params twice?

delete-components/delete-component.sh
```diff
@@ -0,0 +1,108 @@
...
+
+for devenv in dev test ; do
+	oc project ${PROJECT}-${devenv}
+	echo "deleting component environment"
+    oc process cd//component-environment PROJECT=${PROJECT} COMPONENT=${COMPONENT} ENV=${devenv} | oc delete -n ${PROJECT}-${devenv} -f-
+	echo "deleting route"
+    oc process cd//component-route PROJECT=${PROJECT} COMPONENT=${COMPONENT} ENV=${devenv} | oc delete -n ${PROJECT}-${devenv} -f-
+	echo "deleting environment bc"
+    oc process cd//bc-docker PROJECT=${PROJECT} COMPONENT=${COMPONENT} ENV=${devenv} | oc delete -n ${PROJECT}-${devenv} -f-
```
michaelsauter: This will definitely NOT work and can't stay like that. `oc process cd//bc-docker` etc. assume that there is a template stored in OpenShift which there is not.

I would rather use the label `app=$PROJECT-$COMPONENT` and delete all resources with that. That would be the best way forward I think.

kiwo: i will remove the whole shell file and just go with:
oc project <project>
oc stop all -l <component>
oc delete all -l <component>

kiwo: I much prefer this new solution =)

michaelsauter:
@kiwo Please check if you can pass the namespace to `oc delete` ...

PR #319, opened 2019-12-07
start from scratch to reflect openshift

### LINKED ISSUE #318: Components previously deleted reappear in cloned environment.
Openshift resources are exported into the `<project>-occonfig-artifacts` repo during cloning or the daily automated export.

If one removes a component the related files are not deleted in  `<project>-occonfig-artifacts`. As a consequence components previously delete reemerge.

Presumably on export we could simply `rm -rf` the are that was checked out and is written to.

henrjk:
This is the initially suggested fix, but not tested at all.

Once tested should close #318

### DIFF
--- a/ocp-scripts/export-project.sh
+++ b/ocp-scripts/export-project.sh
@@ -253,6 +253,9 @@ echo " -- check out git for $git_checkout_expression"
 eval ${git_checkout_expression}
 echo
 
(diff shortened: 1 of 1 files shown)


michaelsauter:
@henrjk Can you test it today or tomorrow? Would be much appreciated so that we can close this off for ODS 2.

Issue #169, opened 2019-08-09
Use OpenShift OAuth for Nexus


michaelsauter:

kiwo:
What kind of auth is available in Openshift?
Are we connected to pingfederate via oAuth?

michaelsauter:
Sorry I don't know the details here. I think @hugowschneider played with this a while back?

hugowschneider:
The Airflow QS uses openshift OAuth for authentication and authorization like Jenkins.   OpenShift OAuth uses the same authentication backend plunged into openshift. I would advocate for all authentication and authorization inside ODS (like Jenkins, nexus, provisioning and só on) to be plugged into OC OAuth.

PR #328, opened 2019-12-10
Add Oauth config for nexus (#169

kiwo:
**To set this up in Nexus:**

Go to "administration/system/capabilities" in the UI
Click on "new" to add a new capability
Select the "Rut Auth" capability
Fill in: X-Forwarded-User
Save

- Oauth window in front of nexus 
- no need for 2nd tls reencrytion
- cookie secret is not optional :(
- Username Group mapping needs to be done manually or loaded via ldap

ToDo:

- [x] Configure Nexus to use RequestHeaderAuth (and hope params match up) (example https://support.sonatype.com/hc/en-us/articles/214942368-How-to-Configure-Request-Header-Authentication-in-Nexus-with-Apache ?)
- [x] Cleanup Profile for Tailor to allow Crowd and Oauth


nexus/ocp-config/dc.yml
```diff
@@ -26,7 +45,22 @@ objects:
...
+          ports:
+          - containerPort: 8082
+            name: public
+          args:
+          - --https-address=
+          - --http-address=:8082
+          - --provider=openshift
+          - --openshift-service-account=nexus3
+          - --upstream=http://localhost:8081
+          - --cookie-secret=SECRET
```
kiwo: Todo: Gererate random secret ( copy solution from ungerts?)

nexus/ocp-config/dc.yml
```diff
@@ -26,7 +45,22 @@ objects:
...
+        serviceAccountName: nexus3
         containers:
+        - name: oauth-proxy
+          image: openshift/oauth-proxy:latest
+          imagePullPolicy: IfNotPresent
+          ports:
+          - containerPort: 8082
+            name: public
+          args:
+          - --https-address=
```
kiwo: this stops oauth-proxy from loading tls stuff.
in this pr ssl termination is only done in openshift

stefanlack:
Close this, becaus it will be replaced by https://github.com/opendevstack/ods-core/pull/389

PR #329, opened 2019-12-11
Forward env params to build

### LINKED ISSUE #237: Build endpoint does not update env params
When passing environment variables to the `/build` endpoint, the webhook proxy creates a pipeline with those variables. Further invocations via the webook proxy will use the values from the initial creation - independent if you pass any env variables, or pass different values. In Jenkins itself, it is possible though to start the job with different values.

According to https://docs.openshift.com/container-platform/3.11/dev_guide/builds/triggering_builds.html#generic-webhooks it is possible to pass new variables when triggering a build, and if the provided vars collide with existing ones, the values of the passed ones take precedence.

I think that the current behaviour of the webhook proxy is surprising / buggy and that it should take the env variables from the `/build` request and put them into the trigger request to OpenShift.

Before doing that change, it would be nice if @metmajer and @kiwo could comment if that makes sense for you as well or if you would want to see a different behaviour ...

I am not entirely sure that we want to go down that route as the webhook proxy initially was just that, a proxy. So it had to do create+trigger in one invocation. One could argue that the current `/build` endpoint should really be two invocations: create and, separately, trigger ... however doing that would break existing consumers of `/build`.

michaelsauter:
As a consequence, env params set when initially creating the pipeline
might be overriden by further calls.

Closes #237.


michaelsauter:
@henrjk @clemensutschig I fixed the previous linter error - this should be ready for review now.

jenkins/webhook-proxy/main.go
```diff
@@ -428,11 +428,18 @@ func (c *ocClient) Forward(e *Event, triggerSecret string) ([]byte, error) {
...
+	}{
+		Env: e.Env,
+	}
+	b := new(bytes.Buffer)
+	err := json.NewEncoder(b).Encode(p)
+	if err != nil {
+		return nil, fmt.Errorf("Could not encode payload: %s", err)
+	}
+
+	req, _ := http.NewRequest("POST", url, b)
```
clemensutschig: @michaelsauter  and that works? updating an existing pipeline with a post?

michaelsauter: Yes. See the original issue. The behaviour is documented at https://docs.openshift.com/container-platform/3.11/dev_guide/builds/triggering_builds.html#generic-webhooks.

It does not update the pipeline object, it sets the params on the build object.

Issue #327, opened 2019-12-09
Jenkins Master version / source


tbugfinder:
**Feature Request**

The Jenkins version of  the registry repo source of Jenkins master seems to be older than "expected" - especially as security fixes were added to later versions.
The image tagged latest is older than the tag v3.11. The docker.io registry repo isn't maintained any longer.
The logfiles show:
```
INFO: Started w.@53499d85{Jenkins v2.150.2,/,file:///var/lib/jenkins/war/,AVAILABLE}{/var/lib/jenkins/war}
```
**Describe the solution you'd like**
Use later Jenkins version for the Jenkins Master, e.g. tag v3.11.

**Describe alternatives you've considered**
jenkins:lts or test new origin images of quay.io.
https://quay.io/repository/openshift/origin-jenkins?tag=latest&tab=tags
```
docker pull quay.io/openshift/origin-jenkins:latest
```
**Additional context**

michaelsauter:
@tbugfinder Are you referring to the tag `2`? This is the base image for the Jenkins master, see https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/bc.yml#L38.

To my knowledge, this is up-to-date, at least it gets updates quite frequently.

The base image for the slave can be configured, see https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/bc.yml#L79. We use the `3.11` tag internally, as the `latest` tag seems outdated.

tbugfinder:
I didn't check the `bc.yml`.
I run build locally using docker + Dockerfile. So this file is kind of out-of-sync:
https://github.com/opendevstack/ods-core/blob/master/jenkins/master/Dockerfile#L1

michaelsauter:
Yes it should at least say that the value has no effect on OpenShift as the BuildConfig wins ...

tbugfinder:
I'll add a PR :-)
Which Jenkins version is expected for ODS1.2.x and ODS2 deployments?

michaelsauter:
Great!

Latest of v2 of Jenkins. OpenShift version we cannot assume, anything from 3.9 to 3.11 should be possible and up to the user to configure.

clemensutschig:
This is fixed - and controllable in ods-master.
https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L136

PR #332, opened 2019-12-12
Add comment to Dockerfile, use specific tag for FROM

tbugfinder:
Referring to #327 a comment is added to Dockerfiles and the latest v3.11 tag is applied for builds using FROM docker hub.
During test-build scripts were not executed, therefore executable bit was set explicitly.


jenkins/master/Dockerfile
```diff
@@ -11,16 +12,19 @@ RUN yum -y install openssl gnutls-utils \
 
 # fetch certificates and store them in tmp directory
 COPY ./import_certs.sh /usr/local/bin/import_certs.sh
-RUN import_certs.sh
+RUN chmod o+x /usr/local/bin/import_certs.sh \
```
gerardcl: hi! have you tried `sh import_certs.sh` instead of having to change permission set?

gerardcl:
when moving files around permissions might change (and, I agree, fixing the permission set before executing is an option), but it is recommended to stick to execution permission at user level as much as possible via `sh script.sh` instead of fixing, for example, execution rights forever.

jenkins/master/Dockerfile
tbugfinder: I don't have a preference, yet.
If permissions are not set as needed, those might be wrong any way.
This also means that `sh import_certs.sh` could execute successfully, but permissions of `import_certs.sh` might not be as expected.

tbugfinder:
My understanding is that build should pull latest image of `openshift/jenkins-slave-base-centos7`.  Even if ODS is running on OCP3.6, the image tagged `:latest` would be pulled.
The issue in this case is, that tag `:latest` doesn't refer to latest image as it is not properly maintained (https://github.com/openshift/jenkins) and is being migrated to [quai.io](https://quay.io/repository/openshift/origin-jenkins?tag=latest&tab=tags) without support of OCP3.x.

michaelsauter:
@tbugfinder Then maybe best it to make it configurable ... that we show that you should set this to a tag corresponding to the OpenShift version used.

Regarding file permissions, I prefer to fix the bits in the Dockerfile. We have the control there, so I think this is better ... but hey, both works.

tbugfinder:
@michaelsauter How about `ORIGIN_IMAGE_TAG=latest|v3.11|4.4` (derived from ODS_IMAGE_TAG) ?

michaelsauter:
@tbugfinder Could work. I am not sure whether you would need different tags based on which image you use?! Also, maybe we want to allow configuring the whole image, not just the tag?

jenkins/slave-base/Dockerfile.centos7
```diff
@@ -1,4 +1,7 @@
-FROM openshift/jenkins-slave-base-centos7
+# BuildConfig (../ocp-config/bc.yml) overrides FROM instruction
```
michaelsauter: Shouldn't we add the config there then?

tbugfinder: yes, partially it's already there.
https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/bc.yml#L121-L124

michaelsauter: Yes sorry, I should have commented for the master image. The slave image config is already there. I would have actually preferred if your proposed solution would also be just one variable for both image and tag - I don't see the value of having both?

hugowschneider:
Hi guys, I just notice the tests picked up the first PR error:

```
Cloning "http://172.17.0.1:8080/scm/opendevstack/ods-core.git" ...
	Commit:	53bfca04109115fde9ba8d32b48980e197cc6e0b (Merge 017e9801635268c47af53ed449ae0887ad6831e8 into c89cc4c33ac466ed848b390cfa..)
	Author:	tbugfinder <github@online.ms>
	Date:	Sun Jan 12 20:39:28 2020 +0000
Replaced Dockerfile FROM image ${JENKINS_AGENT_BASE_FROM_IMAGE}

Pulling image openshift/jenkins-slave-base-centos7 ...
Pulled 1/7 layers, 17% complete
Pulled 2/7 layers, 30% complete
Pulled 3/7 layers, 52% complete
Pulled 4/7 layers, 65% complete
Pulled 5/7 layers, 73% complete
Pulled 6/7 layers, 88% complete
Pulled 7/7 layers, 100% complete
Extracting
error: build error: no FROM statement found
```

Tests are working 👍

tbugfinder:
> Hi guys, I just notice the tests picked up the first PR error:
> 
> ```
> Cloning "http://172.17.0.1:8080/scm/opendevstack/ods-core.git" ...
> 	Commit:	53bfca04109115fde9ba8d32b48980e197cc6e0b (Merge 017e9801635268c47af53ed449ae0887ad6831e8 into c89cc4c33ac466ed848b390cfa..)
> 	Author:	tbugfinder <github@online.ms>
> 	Date:	Sun Jan 12 20:39:28 2020 +0000
> Replaced Dockerfile FROM image ${JENKINS_AGENT_BASE_FROM_IMAGE}
> 
> Pulling image openshift/jenkins-slave-base-centos7 ...
> Pulled 1/7 layers, 17% complete
> Pulled 2/7 layers, 30% complete
> Pulled 3/7 layers, 52% complete
> Pulled 4/7 layers, 65% complete
> Pulled 5/7 layers, 73% complete
> Pulled 6/7 layers, 88% complete
> Pulled 7/7 layers, 100% complete
> Extracting
> error: build error: no FROM statement found
> ```
> 
> Tests are working 👍

Nice.
@hugowschneider  Which docker version is running in the environment?

michaelsauter:
@tbugfinder This is inside OpenShift, so whatever version 3.11 uses (I think the Docker version there is quite old)

tbugfinder:
Well, 3 years old docker 1.13.1 (2017-02-08) doesn't support ARG within FROM. So using FROM or a tag based on an input variable isn't feasible. Still the root issue is, that tag `:latest` isn't a valid image tag for a v3.11 installation.

Ref:
- https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html
- https://docs.docker.com/release-notes/docker-engine/

tbugfinder:
I'm closing the PR as the version of the underlying docker installation is too old to support the required features.

Issue #336, opened 2019-12-16
create-projects: Bug during argument parsing


hugowschneider:
the option -d/--debug should not have shift:

https://github.com/opendevstack/ods-core/blob/4b51909ee622406dbd9c37f9836ca54a684569b4/create-projects/create-cd-jenkins.sh#L24-L26

Issue #335, opened 2019-12-16
create-projects: Using tailor --force by default in create-cd-jenkins.sh


hugowschneider:
`tailor --force` should not be a default option inside a script and it is in a variable with misleading name

https://github.com/opendevstack/ods-core/blob/4b51909ee622406dbd9c37f9836ca54a684569b4/create-projects/create-cd-jenkins.sh#L42

Issue #331, opened 2019-12-12
create-cd-jenkins.sh: `type -P` does not work in all linux distros


hugowschneider:
https://github.com/opendevstack/ods-core/blob/0bbb48417085d15515c732fe45185581f74f9c1c/create-projects/create-cd-jenkins.sh#L9

does not work in all linux distros. Testing the script in some different environments like:

- Ubuntu Xenial in travis: https://travis-ci.com/opendevstack/ods-core/jobs/266490723#L630
```
+type -P tailor
+tailor_exe=-P: not found
tailor is /usr/local/bin/tailor
```
- MacOS
`
type: bad option: -P
`

henrjk:
While I think this can be easily adjusted in this case. There could be other cases where it is more complicated to ensure wider compatibility.

Also on my mac `type -P` works:

```
~ $ type -P tailor
/usr/local/bin/tailor
~ $ type -P type
/usr/bin/type
```

For this script I assumed that compatibility with cd/jenkins-slave-base would be required.
Macos compatibility would be nice to have for sure as many opendevstack devs are mac users.

Nonetheless I am wondering what requirements there are regarding distro compatibility?

hugowschneider:
It need to run in travis, that is the major pain point

henrjk:
I did some research and found that `type -P` should be available on travis xenial version.
@hugowschneider and I verified this earlier today. 

Now it is less clear why this is not available when the command is executed using the following function [RunScriptFromBaseDir](https://github.com/BIX-Digital/ods-core/blob/eeff9495aab9afc23ef9b49044a12e1b3f8cd18e/tests/utils/command-utils.go#L15). 

I found a reference that go would intentionally not use the system shell (see https://stackoverflow.com/a/52281031). 

Inspired from this I also noted on a ubuntu 1804 virtual box machine:

```bash
~$ bash -ctype -P ls
bash: -y: invalid option
```
whereas

```bash
~$ bash -c type -P ls
~$ echo $?
0
```

PR #338, opened 2019-12-16
Removing type -P and --force from script

hugowschneider:
This PR refers to the issues https://github.com/opendevstack/ods-core/issues/331, https://github.com/opendevstack/ods-core/issues/335 and https://github.com/opendevstack/ods-core/issues/336

In the script https://github.com/opendevstack/ods-core/blob/master/create-projects/create-cd-jenkins.sh:

- It removes the `type -P`
- Refactor the command line parser adding usage 
- Remove the --force from the script and adds it as a argument to the script
- Add in the Jenkins file --force to keep compatibility

\closes https://github.com/opendevstack/ods-core/issues/331
\closes https://github.com/opendevstack/ods-core/issues/335
\closes https://github.com/opendevstack/ods-core/issues/336

### DIFF
--- a/create-projects/create-cd-jenkins.sh
+++ b/create-projects/create-cd-jenkins.sh
@@ -1,80 +1,82 @@
 #!/usr/bin/env bash
-set -eux
+set -eu
 
 # This script sets up the cd/jenkins-master and the associated webhook proxy.
 
 # support pointing to patched tailor using TAILOR environment variable
-: ${TAILOR:=tailor}
 
-tailor_exe=$(type -P ${TAILOR})
-tailor_version=$(${TAILOR} version)
-
-echo "Using tailor ${tailor_version} from ${tailor_exe}"
-
+TAILOR="tailor"
 DEBUG=false
 STATUS=false
-while [[ $# -gt 0 ]]
-do
-key="$1"
+FORCE=""
 
-case $key in
-    --status)
-    STATUS=true
-    ;;
-    -d|--debug)
-    DEBUG=true;
-    shift
-    ;;
-    *)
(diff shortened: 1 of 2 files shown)


create-projects/create-cd-jenkins.sh
```diff
@@ -1,80 +1,82 @@
...
+   printf "\t-h|--help\tPrints the usage\n"
+   printf "\t-d|--debug\tEnabled debug\n"
+   printf "\t-t|--tailor\tChanges the executable of tailor. Default: tailor\n"
+
+}
+
 
+while [[ "$#" -gt 0 ]]; do case $1 in
+   --status) STATUS=true;;
+   -d|--debug) DEBUG=true; set -x;;
```
michaelsauter: I like it, but that was the default previously. Maybe we add it to the arguments in the Jenkinsfile? Also, I would prefer this to be called `--verbose` to align with Tailor

create-projects/create-cd-jenkins.sh
```diff
@@ -1,80 +1,82 @@
...
+   --force) FORCE="--force"; ;;
+   -h|--help) usage; exit 0;;
+   -t=*|--tailor=*) TAILOR="${1#*=}";;
+   -t|--tailor) TAILOR="$2"; shift;;
```
michaelsauter: Can we separate each group of args by a newline? Think that would make it even clearer ...

hugowschneider: Done :)

create-projects/create-cd-jenkins.sh
hugowschneider: Done!

PR #342, opened 2019-12-17
Initial ODS-CORE Test Infrastructure -  Merry Christmas

hugowschneider:
This is the initial work of setting up a test infrastructure to ods-core. This PR includes:

- tests: Folder containing all tests. This folder should mimic the root folder structure
- ods-setup: Folder containing scripts to setup ODS. This is an attempt to split the scripts from infrastructure setup into smaller and simple scripts. For now, it is being used for the test but it could be used to deploy ODS.


metmajer:
@hugowschneider very interesting indeed. We are working on setting up ODS development environments and since the local installation documentation is not complete, I guess it would make sense to look into this. Thoughts?

hugowschneider:
@metmajer, I guess when we start implementing comprehensive tests, the local installation will be clearer and in a point to be documented. I am still exploring the local installation with `oc cluster up` what brings some pitfalls (for example BitBucket). 

Currently, I have a VM with a replica of the TravisCI test environment and it is working so far. The challenges will come when I start mocking/replicating the Atlassian stack.

metmajer:
@hugowschneider seems like we are having the same thoughts. I had seen the development environment as the first step to having an integration testing environment ready. What is your plan for the Atlassian stack? Run Jira and Bitbucket in Docker?

hugowschneider:
@metmajer It is not possible to run the Atlassian stack on a Docker because it needs a license and the license is bound to an account, which means it will not work.

I also looked into having an Atlassian Cloud account and use TravisCI encryption for storing the credentials, but this TravisCI feature does not work for PRs coming from forks (https://docs.travis-ci.com/user/encryption-keys/). In my opinion, it will have a huge impact when people contribute to the source. I don't see another way of storing credentials.

Last resort is what I am doing now. I am creating an image that is a git server with some of the APIs mocked.

metmajer:
@hugowschneider apologies, but this does work. We have been using this locally for a while with a trial license that expires after one week. Not a long-term solution, but I can live with a solution where we'd need to apply a license manually.

hugowschneider:
@metmajer I meant it does not work because once the license expires the builds will get blocked, for sure with the trial it works. Also, the problem is how do we bypass the installation steps? And I do not know if I can request for trial licenses every time one expires. And in the end, I cannot safely store the license without blocking builds coming from forks... Do you have any ideas?

tjaeschke:
@hugowschneider If you archive expired trial licenses you are not limited to a fixed amount, but like you said, bypassing the installation steps is not as simple as it should be in my opinion. Only Bitbucket provides unattended installation. You can bypass the installation steps, if you use database backups or automate the steps with mechanize or selenium.

hugowschneider:
I do not know what will be less painful, mocking or automate the whole installation inside TravisCI. I for me safely storing the license is still an open question.

michaelsauter:
We'll get huge benefits by being able to test project and quickstarter provisioning. There are four parts that touch Bitbucket as far as I can tell:

1. Create repo & webhook from prov-app
2. Push to repo
3. Clone from repo
4. Set build status from shared library

Depending on how tricky Bitbucket proves to be to install ad-hoc, I think it would be better to make progress with the tests. @hugowschneider and I discussed this and think that using https://gogs.io might be an option. I'd rather be able to test the rest and not be blocked / held up by Bitbucket. Not perfect, but a good first step.

tjaeschke:
@rattermeyer, @stefanlack and I are on the same track. Richard recommended https://gitea.io/, but I think this is a detail question.

metmajer:
@hugowschneider the license activation is really a bummer to create an ad-hoc environment :(

- I'll investigate if there's something one can do if you at least have an account created already. 
- Going public cloud is an option, but for my use-case, not a feasible way forward.
- Also, the _release manager_ needs to be integration tested against Bitbucket, Jira and Nexus.

hugowschneider:
I am also researching which setup will be the lighter for an ad-hoc environment. All the tools so far have installations steps which I need to figure out how to bypass.

henrjk:
@hugowschneider this is very cool. Thanks for pushing this out here. I hope I will have the cycles available for a review in the next days.

tests/create-projects/create-projects_test.go
```diff
@@ -0,0 +1,99 @@
...
+    if err = utils.FindRoleBinding(roleBindings, "default", "ServiceAccount", utils.PROJECT_NAME_TEST, "system:image-builder"); err != nil {
+        t.Fatal(err)
+    }
+
+    if err = utils.FindRoleBinding(roleBindings, "jenkins", "ServiceAccount", utils.PROJECT_NAME_CD, "admin"); err != nil {
+        t.Fatal(err)
+    }
+
+    t.Log("WARNING: Seeding special and default permission groups is not tested yet!")
+
```
clemensutschig: @hugowschneider  - where do we do the testing if e.g. jenkins / webhook proxy are running .. because then we really know that the seed of a project works ... :)

hugowschneider: The webhooks and git server integration test will come with "MockBucket" (https://github.com/opendevstack/mockbucket). A service mock that integrates git server, webhooks to the test suite of each quickstarter (yet to come)... :) 

I will also do a PoC on how to run Jenkins pipelines within the local oc cluster.

clemensutschig: @hugowschneider  - for the base jenkins check no need for the mockbucket ... or?

hugowschneider: For the deployment test that I did so far no... the MockBucket will be usefull when we need to test git commits/pushes and trigger builds

.travis.yml
```diff
@@ -2,16 +2,42 @@
...
+                - wget https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz
+                - tar -xzvf openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz
+                - export PATH=$PATH:$PWD/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit
+                - "sudo cat /etc/docker/daemon.json | jq '. + {\"insecure-registries\": [\"172.30.0.0/16\"]}' | sudo tee /etc/docker/daemon.json"
+                - sudo cat /etc/docker/daemon.json
+                - sudo service docker restart
+                - curl -LO "https://github.com/opendevstack/tailor/releases/download/v0.11.0/tailor-linux-amd64"
+                - chmod +x tailor-linux-amd64
+                - sudo mv tailor-linux-amd64 /usr/local/bin/tailor
+                - oc cluster up --base-dir=${HOME}/openshift.local.clusterup
```
henrjk: `oc cluster up` appears to be not available under openshift 4 which is mentioned at [oc cluster not working on mac anymore · Issue #21512 · openshift/origin](https://github.com/openshift/origin/issues/21512)

tests/utils/command-utils.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"runtime"
+)
+
+func RunScriptFromBaseDir(command string, args []string, envVars ...string) (string, string, error) {
+	_, filename, _, _ := runtime.Caller(0)
+	dir := path.Join(path.Dir(filename), "..", "..")
+	return RunCommand("bash", append([]string{fmt.Sprintf("%s/%s", dir, command)}, args...), envVars...)
```
henrjk: Perhaps a '-c' in here (see https://github.com/opendevstack/ods-core/issues/331#issuecomment-566449448) would allow for type -P as expected?

metmajer:
@henrjk it is in fact not available with OpenShift 4. CodeReady containers will be the way to go for a local OpenShift v4 clusters and beyond: https://github.com/code-ready/crc.

hugowschneider:
@henrjk 
> I left two other comments with a minor observation and an issue I came across mentioning `oc cluster up` may not be available on OpenShift 4. Do you have further insights about this?

Like @metmajer said, CodeReady containers are the way to go, but I am definitely not looking into it since I believe there will be more changes needed in the code. 

> Another question that came to mind (perhaps because of my ignorance) is why both k8s client-go and openshift's client-go are used?

The k8s client-go is a dependency of the Openshift client-go :)

.travis.yml
hugowschneider: Currently, because of bugs, hopefully, only because of bugs, `oc cluster up` does not work on macs. Also, I am not considering Openshift 4 since I'm not sure if ODS is supporting it, but tailor is not, and tailor is needed.

tests/utils/command-utils.go
hugowschneider: I would favor the PR #338 and disregard -P for now, but also these scripts need `chmod +x`. There is a lot of improvements needed which could go as a separate issue.

.travis.yml
michaelsauter: Correct. ODS has not been tested against OpenShift 4. Tailor does not work with it right now, but it does work with https://github.com/opendevstack/tailor/pull/160. OpenShift 4 can be run locally using https://github.com/code-ready/crc.

hugowschneider:
I am closing this PR in favor of #356

PR #348, opened 2019-12-27
in clone-project.sh use export, import from SCRIPT_DIR

henrjk:
This change is made in anticipation of opendevstack/ods-jenkins-shared-library#167 which will remove the responsibility for the clone script to curl its own dependencies. 

Instead it uses the relative scripts which are expected to be in the same relative location as if one checks out this repository. 

Only when this is merged the branch configuration to be provided in a PR for the ods-jenkins-shared-library will work.

### DIFF
--- a/ocp-scripts/clone-project.sh
+++ b/ocp-scripts/clone-project.sh
@@ -88,14 +88,13 @@ echo "Provided params: \
 SOURCE_PROJECT="$PROJECT_ID-$SOURCE_ENV"
 TARGET_PROJECT="$PROJECT_ID-$TARGET_ENV"
 
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 echo "[INFO]: creating workplace: mkdir -p oc_migration_scripts/migration_config"
 mkdir -p oc_migration_scripts/migration_config
 cd oc_migration_scripts
 echo $(pwd)
-export_url="https://$BITBUCKET_HOST/projects/opendevstack/repos/ods-core/raw/ocp-scripts/export-project.sh?at=refs%2Fheads%2Fproduction"
-curl --fail -s --user $CREDENTIALS -G $export_url -d raw -o export.sh
-import_url="https://$BITBUCKET_HOST/projects/opendevstack/repos/ods-core/raw/ocp-scripts/import-project.sh?at=refs%2Fheads%2Fproduction"
-curl --fail -s --user $CREDENTIALS -G $import_url -d raw -o import.sh
+cp -v "$SCRIPT_DIR/export-project.sh" .
+cp -v "$SCRIPT_DIR/import-project.sh" .
 
 cd migration_config
 echo $(pwd)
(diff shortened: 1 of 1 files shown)


clemensutschig:
As the scripts have been renamed.. was this Working with ods2?

henrjk:
I don't see issues with the renaming inside this PR as this should be an internal detail.I made the change so one finds this if one greps for the file name. 

However the ODS 2 branch curls the clone script from the production branch at  https://github.com/opendevstack/ods-jenkins-shared-library/blob/a7ff00a69518c447829aa00741cdd8d849bb9c8e/src/org/ods/OdsPipeline.groovy#L270 and expect the clone script to curl the dependent scripts which with this change would no longer do. 

The ODS 2 branch should probably not refer to production for the curl. Potentially this could go into ODS 2 as a bug fix. On the other hand other branches still in use would also need this fix such as 1.2.x. 

A backward compatible alternative would be to not touch the original clone-project.sh but to deprecate it. The changed behavior would be introduced in a script named `clone-project-nocurl.sh` and  opendevstack/ods-jenkins-shared-library#167  would be adjusted to use that script instead.

clemensutschig:
I dont See the rename though..  kg was this merged earlier..?

clemensutschig:
Got it. Just Merge.

henrjk:
The backwards compatibility concerns discussed above make me think that pursuing the proposed backwards-compatible alternative above is better.

For example I know that there are several projects which have changed to use the shared library from `ods-library@1.2.x`. For these autocloning would break if we merged as is.

henrjk:
I pushed branches which show the backwards compatible approach to BIX-Digital:

- https://github.com/BIX-Digital/ods-core/tree/feature/jsl-167-improve-clone-script-dev-bwc
- https://github.com/BIX-Digital/ods-jenkins-shared-library/tree/feature/jsl-167-improve-clone-script-dev-bwc

michaelsauter:
@henrjk I see your point. I think right now not everything has fully transitioned to an optional production branch, and the best way to install ODS would be to have the production branch in sync with the branch you have "installed". 

One alternative to having two scripts would be to check if the files exist, and if not, curl them as before. I think I'd prefer this solution because having two scripts will likely cause confusion: people seeing the scripts without the context here might ask why they exist. Further, we need to potentially maintain both and it is unclear at which point we'd remove the "curl" version, and then we are left with a "nocurl" version .... :/

What do you think?

ocp-scripts/clone-project.sh
```diff
@@ -88,14 +88,13 @@ echo "Provided params: \
...
 echo "[INFO]: creating workplace: mkdir -p oc_migration_scripts/migration_config"
 mkdir -p oc_migration_scripts/migration_config
 cd oc_migration_scripts
 echo $(pwd)
-export_url="https://$BITBUCKET_HOST/projects/opendevstack/repos/ods-core/raw/ocp-scripts/export-project.sh?at=refs%2Fheads%2Fproduction"
-curl --fail -s --user $CREDENTIALS -G $export_url -d raw -o export.sh
-import_url="https://$BITBUCKET_HOST/projects/opendevstack/repos/ods-core/raw/ocp-scripts/import-project.sh?at=refs%2Fheads%2Fproduction"
-curl --fail -s --user $CREDENTIALS -G $import_url -d raw -o import.sh
+cp -v "$SCRIPT_DIR/export-project.sh" .
+cp -v "$SCRIPT_DIR/import-project.sh" .
```
michaelsauter: I think these can be removed. I only see two ways this script can be used:

1. It is curled. Then the other files are also curled and all files are in the same folder. Just calling `sh export-project.sh` will work.

2. It is not curled, but cloned via Git. Then the other files are in the same folder as well, which means just calling `sh export-project.sh` will work too.

henrjk: I like your proposed solution with the clone-script checking itself for the presence of export import scripts and curling as needed.

henrjk: I tested this version in our environment including for backwards compatibility.

PR #351, opened 2020-01-02
Adding port option to APP_DNS for fetching certificates

### LINKED ISSUE #350: Import Certificates during Jenkins build
During import_certs.sh in the line:

https://github.com/opendevstack/ods-core/blob/04bfd8ce04466263a5440572d66563c460fd2904/jenkins/master/import_certs.sh#L14

Also in:

https://github.com/opendevstack/ods-core/blob/04bfd8ce04466263a5440572d66563c460fd2904/jenkins/slave-base/Dockerfile.centos7#L31-L32

Since there is no port option, it is not possible to build Jenkins in a local cluster, or in a cluster that does not use the default (443) port.  

Also, those lines are duplicated across multiple files:

https://github.com/opendevstack/ods-core/blob/master/jenkins/slave-base/Dockerfile.centos7
https://github.com/opendevstack/ods-core/blob/master/jenkins/slave-base/Dockerfile.rhel7
https://github.com/opendevstack/ods-core/blob/master/jenkins/master/import_certs.sh

A would suggest adding a port variable in the env file.

hugowschneider:
\closes #350 

Added APP_DNS_PORT to all calls of `gnutls-cli` during Jenkins build and set the default to 443

### DIFF
--- a/jenkins/master/import_certs.sh
+++ b/jenkins/master/import_certs.sh
@@ -4,6 +4,8 @@ oldIFS=$IFS
 IFS=';'
 KEYSTORE="$JAVA_HOME/lib/security/cacerts"
 
+: "${APP_DNS_PORT:=443}"
+
 if [ "${TARGET_HOSTS}x" == "x" ] ; then
     TARGET_HOSTS=${APP_DNS}
 fi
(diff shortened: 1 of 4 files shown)


michaelsauter:
@hugowschneider I would like to see a follow-up change that adds this setting to the configuration sample. At least it should be there in a commented form, but I would actually prefer to have the default in the config.

Issue #352, opened 2020-01-06
default config for keycloak


stefanlack:
**Is your feature request related to a problem? Please describe.**
Vagrant box idmanager has a plain keycloak installation. However, no opendevstack specific configuration is available and has to be created´.

**Describe the solution you'd like**
Install default OpenDevStack-admin and OpenDevStack-User and appropriate keycloak client and roles, so that no additional manual configuration is required.

PR #353, opened 2020-01-06
#352 default config for keycloak

stefanlack:


stefanlack:
Do not merge. I need to update this. We are using a protocoll mapper that mapps group memberschip to id token claims. Therefore, we have to create groups and add group membershipt to default user admin1 and user1.

Alternativ solution would be to add an role mapper

stefanlack:
Implemented group membership.

PR #354, opened 2020-01-06
Provide clone-project-no-curl.sh which expect export/import in same dir.

henrjk:
Adds deprecation warning to clone-project.sh

Closes opendevstack/ods-jenkins-shared-library#167 with PR opendevstack/ods-jenkins-shared-library#178


henrjk:
I am retracting this as I am now working on having compatibility code added to the original clone-script.

PR #370, opened 2020-01-13
Run proxy tests via GitHub Actions

michaelsauter:
The build failed previously because Travis tried to download the modules specified in the `tests` directory as well - it should have just fetched modules needed for the webhook proxy.

This PR fixes it by just moving to GH, which works correctly out of the box.

Also, I'm telling the workflow to run more often - not sure why you had the restriction previously @hugowschneider?


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -1,11 +1,40 @@
...
-    name: ODS resources setup and provisioning tests
+  webhook-proxy:
+    name: Webhook Proxy tests
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: Checkout repository
+        uses: actions/checkout@v2.0.0
+        with:
+          fetch-depth: 0
```
hugowschneider: You can use default fetch-depth (or fetch-depth 1) if you do not need the whole git history. It is faster

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -1,11 +1,40 @@
...
-  pull_request:
-    branches:
-      - master
+on: [push, pull_request]
 jobs:
-  test:
-    name: ODS resources setup and provisioning tests
+  webhook-proxy:
+    name: Webhook Proxy tests
+    runs-on: ubuntu-18.04
```
hugowschneider: As a question... which version should we adopt 18.04 or 16.04? I used 16.04 just because it was the default for TravisCI

hugowschneider:
I love the green ✔️ 😃

.github/workflows/continuous-integration-workflow.yml
michaelsauter: Why stick with an older version? For me this should be latest as much as possible. I'm only avoid using `latest` because then the actual version is undefined.

hugowschneider: I agree... should we change also the other test then to 18.04? Just to make it consistent

hugowschneider: Or create a issue/PR for it maybe

.github/workflows/continuous-integration-workflow.yml
hugowschneider: 👍

.github/workflows/continuous-integration-workflow.yml
hugowschneider: 👍

PR #394, opened 2020-01-20
Checking if Jenkins images are being successfully built.

### LINKED ISSUE #393: Test should fail when image build fails
This lines should fail when build fails

https://github.com/opendevstack/ods-core/blob/802ebc65c7e9f04b81d5ab9261db64d94403aae2/ods-setup/setup-jenkins-images.sh#L68-L70

Otherwise, we get test errors which are hard to correlate back to image build errors

hugowschneider:
/closes #393 

Since `oc start-build --follow` does not provide any exit code if the build fails, the code now checks after each build if the status phase of it is Complete.

### DIFF
--- a/ods-setup/setup-jenkins-images.sh
+++ b/ods-setup/setup-jenkins-images.sh
@@ -66,5 +66,20 @@ ${TAILOR} update ${FORCE} --context-dir=${BASH_SOURCE%/*}/../jenkins/ocp-config 
 
 echo "Start Jenkins Builds"
 oc start-build -n ${NAMESPACE} jenkins-master --follow
+STATUS=$(oc get build jenkins-master-1 -o jsonpath='{.status.phase}')
+if [ "${STATUS}" != "Complete" ]; then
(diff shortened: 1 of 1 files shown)


ods-setup/setup-jenkins-images.sh
```diff
@@ -66,5 +66,20 @@ ${TAILOR} update ${FORCE} --context-dir=${BASH_SOURCE%/*}/../jenkins/ocp-config
 
 echo "Start Jenkins Builds"
 oc start-build -n ${NAMESPACE} jenkins-master --follow
+STATUS=$(oc get build jenkins-master-1 -o json -n cd | jq '.status.phase' | sed 's/"//g')
```
michaelsauter: This is easier like this: https://github.com/opendevstack/ods-jenkins-shared-library/blob/master/vars/stageStartOpenshiftBuild.groovy#L33

hugowschneider: 👍

Issue #381, opened 2020-01-17
Create Makefile to give easy access to common tasks


michaelsauter:
**Is your feature request related to a problem? Please describe.**
Right now it is difficult to find which "commands" are available, and how to do certain tasks such as updating SonarQube.

**Describe the solution you'd like**
I would like to have a `Makefile` that contains various targets which make it easy to run commands. Examples would be:

* `make jenkins`: Run `tailor update` on Jenkins build resources and start/follow builds
* `make sonarqube`: Run `tailor update` on SQ build resources (build configs etc.), build the image, run `tailor update` on SQ run resources (deployment configs etc.)
* `make nexus`: Run `tailor update` on Nexus resources (no custom image like SQ, so easier to handle ...). Later on that could handle repository setup in Nexus too.
* `make cd-setup`: Creates / updates resources in the CD namespace like cd-user, rolebindings
* `make configuration`: Runs `configuration-sample/update` script

**Describe alternatives you've considered**
None.

**Additional context**
This would pave the way to an easier installation / update procedure.

@metmajer I think you mentioned sth. like this before as well. I can't find it right now though. Did you have in mind what I described or different ideas?

@hugowschneider This would make use of some of the scripts you wrote ... and on the other hand, the tests could use this `Makefile` to do the setup. What do you think?

metmajer:
@michaelsauter will see if I still have the links... will provide an update today

metmajer:
@michaelsauter I couldn't find these issues quickly either, but what I had in mind goes a bit beyond:

`make test`: runs automated tests -- which could be scoped to the current repo or, if invoked in some meta-(test-)repo, would exhaustively test the entire ODS

`make distribute`: to distribute an ODS codebase from a configurable upstream repo to a configurable number of downstream repos. Thoughts here are to automate release branch generation, git pull $upstream and tagging. In case of merge conflicts, the user would need to resolve conflicts and re-run (so idempotent behavior would be needed).

`make release`: distribute + whatever else is needed

`make install`: to install or upgrade ODS on a configurable set of clusters. This will require us to define a generic install or upgrade procedure first, but would be an incredible improvement to have.

Would love to elaborate thoughts on this with the team soon.

michaelsauter:
I see, and it sounds good to me. Unfortunately, I think apart from `test`, we are far off from implementing them. Therefore, I suggest to start with what I have built, and extend to cover what you have in mind as soon as we have the building blocks for it.

metmajer:
[...]

PR #395, opened 2020-01-21
Add Makefile for common tasks

michaelsauter:
Addresses #381.

Could be extended for other purposes, but this might be a good start.

Use `make help` to get an overview of available targets.

@gerardcl Maybe we can test-drive this next week? :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ jenkins-start-build-webhook-proxy:
 
 # SONARQUBE
 ## Install or update SonarQube.
-sonarqube: sonarqube-backup sonarqube-apply-build sonarqube-start-build sonarqube-apply-run
+sonarqube: sonarqube-apply-build sonarqube-start-build sonarqube-apply-run
 .PHONY: sonarqube
 
 ## Update OpenShift resources related to the SonarQube image.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,112 @@
+SHELL = /bin/bash
+.SHELLFLAGS := -eu -o pipefail -c
+.DELETE_ON_ERROR:
+MAKEFLAGS += --warn-undefined-variables
+MAKEFLAGS += --no-builtin-rules
```
henrjk: The values look reasonable to me. Do you have a specific source of inspiration for them?

Makefile
```diff
@@ -0,0 +1,112 @@
+SHELL = /bin/bash
```
henrjk: Should this be `SHELL = /usr/bin/env bash` ?

Makefile
```diff
@@ -0,0 +1,112 @@
...
+	tailor export -n cd > backup_cd.yml
+.PHONY: backup-ocp-config
+
+## Create a backup of the SonarQube database in the current directory.
+backup-sonarqube:
+	cd sonarqube && sh backup.sh .
+.PHONY: backup-sonarqube
+
+# HELP
+# Based on https://gist.github.com/prwhite/8168133#gistcomment-2278355.
```
henrjk: Cool to generate the help!

Makefile
michaelsauter: Takne from https://tech.davis-hansson.com/p/make/. It has a few other things that I don't think are helpful as e.g. they need a newer make version than available in MacOS Catalina by default.

Makefile
michaelsauter: Maybe - I never really care that much. Re-reading https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my, your suggestion might be better for portability. Did you ever encounter issues with scripts using just `/bin/bash`?

metmajer: Personally, not with Bash. I have mainly seen this useful when using other interpreters, such as ruby or groovy.

PR #410, opened 2020-01-31
Move ODS config to Jenkins master

michaelsauter:
* Moves the ODS configuration file from the slaves to the master.
* Uses the information in the config file to set the default branch of the global libraries (which removes yet another place which needs `production` to be present).
* Replaces existing libraries with the new configuration on boot so that any parameter changes are reflected.
* Cleanup of log messages.

As a next step, I'll read this configuration in the shared library, and place it into the context there. Then all stages have access to the information, which they can use e.g. to determine whether the PR analysis feature is turned on.


michaelsauter:
@gerardcl @clemensutschig ping

PR #415, opened 2020-02-27
Configure serviceAccount to use for webhook proxy instance

### LINKED ISSUE #413: Webhook Proxy does not run under jenkins serviceaccount
**Describe the bug**
The OCP template does not define the `serviceAccount` to use, but it should be `jenkins`, the same account that the Jenkins master instance uses.

**Affected version:**
 - OpenShift: all
 - OpenDevStack: 1.2.x, 2.x

michaelsauter:
Fixes #413


michaelsauter:
I will do more tests to see if we can restrict permissions of the default role after this change ...

PR #417, opened 2020-02-27
Add new target cd-project

### LINKED ISSUE #414: Serviceaccount "deployment" not needed anymore?
In https://github.com/opendevstack/ods-core/blob/master/ods-setup/setup-ods-project.sh#L52-L53, we create a serviceaccount `deployment` and give it very extensive rights.

From what I understand, this was done for Rundeck, and therefore it should not be necessary anymore. At least I don't know anything that uses it ...

Agree @rattermeyer @tjaeschke @gerardcl @hugowschneider?

michaelsauter:
* Exposes creation/setup of central project to ease installation
* Removes unnecessary deployment serviceaccount

Closes #414.

@hugowschneider I noticed that we can merge the scripts that you use in your tests with the makefile targets, but that needs to be a follow-up PR.

### DIFF
--- a/ods-setup/setup-ods-project.sh
+++ b/ods-setup/setup-ods-project.sh
@@ -38,19 +38,17 @@ if ! oc whoami; then
   exit 1
 fi
 
+# Create namespace
 if oc project ${NAMESPACE}; then
   echo "The project '${NAMESPACE}' already exists"
-  exit 1
+else
+  echo "Creating project '${NAMESPACE}' ..."
+  oc new-project ${NAMESPACE} --description="Central ODS namespace with shared resources" --display-name="OpenDevStack"
 fi
-
-oc new-project ${NAMESPACE} --description="Base project holding the templates and the Repositoy Manager" --display-name="OpenDevStack"
 
 # Allow system:authenticated group to pull images from CD namespace
 oc adm policy add-cluster-role-to-group system:image-puller system:authenticated -n ${NAMESPACE}
 oc adm policy add-role-to-group view system:authenticated -n ${NAMESPACE}
 
-oc create sa deployment -n ${NAMESPACE}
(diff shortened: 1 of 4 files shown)


michaelsauter:
@hugowschneider ping

ods-setup/setup-ods-project.sh
```diff
@@ -38,19 +38,17 @@ if ! oc whoami; then
   exit 1
 fi
 
+# Create namespace
 if oc project ${NAMESPACE}; then
   echo "The project '${NAMESPACE}' already exists"
-  exit 1
```
hugowschneider: If the project exists I would exit and maybe use a flag like FORCE... It could avoid big mistakes

michaelsauter: Hmm yes. I was actually coming from the angle: "could this script be run multiple times?". It turns out that all those policies further down are actually idempotent.

Doing it that way it would be possible to eventually have a `make install` which works for install and for update.

But yea ... I know what you mean. Thoughts?

hugowschneider: e.g... I type instead of setup-ods-project.sh -n **cd**, setup-ods-project.sh -n **cds** (just a stupid typo) and imagine cds could be a project with a lot of things... tailoring it would create a huge mess

michaelsauter: Hmm, so the actual issue seems to be if I use it for a namespace other than `cd`. What I'm trying to add here is a Makefile target which does not allow you to choose a specific name anyway. 

So - do we need to allow other namespaces at all? I would vote to remove this feature.

hugowschneider: We could then put just a BIG warning saying do not run the scripts without make, or do it in your own risk :) LOL... that would be fine.

michaelsauter:
@hugowschneider I removed the `--namespace` flags in 91392d3. Now safe enough? :)

PR #418, opened 2020-02-28
adds LOGGING_FILE_PATH prop app property to configuration sample files

stitakis:
relates and completes https://github.com/opendevstack/ods-provisioning-app/issues/350

### DIFF
--- a/configuration-sample/ods-provisioning-app-dev.env.sample
+++ b/configuration-sample/ods-provisioning-app-dev.env.sample
@@ -41,3 +41,6 @@ PROV_APP_LOG_LEVEL_OPENDEVSTACK=DEBUG
 
 # Jenkins pipeline trigger secret (raw)
 PROV_APP_PIPELINE_TRIGGER_SECRET=changeme
+
+# Log files system location
+LOGGING_FILE_PATH=./logs/changeme
--- a/configuration-sample/ods-provisioning-app-dev.env.sample
+++ b/configuration-sample/ods-provisioning-app-dev.env.sample
@@ -43,4 +43,4 @@ PROV_APP_LOG_LEVEL_OPENDEVSTACK=DEBUG
 PROV_APP_PIPELINE_TRIGGER_SECRET=changeme
 
 # Log files system location
-LOGGING_FILE_PATH=./logs/changeme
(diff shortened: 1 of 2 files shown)


configuration-sample/ods-provisioning-app-dev.env.sample
```diff
@@ -41,3 +41,6 @@ PROV_APP_LOG_LEVEL_OPENDEVSTACK=DEBUG
 
 # Jenkins pipeline trigger secret (raw)
 PROV_APP_PIPELINE_TRIGGER_SECRET=changeme
+
+# Log files system location
+LOGGING_FILE_PATH=./logs/changeme
```
michaelsauter: Isn't there a better default that does not necessarily need to be changed?

stitakis: mmm...not sure what you mean... assuming that one of the roles of this example configuration is to highlight the properties that are environment specific and need to be changed, we can put something like ```/opt/changeme```, looks better for you?

stitakis: @michaelsauter did you mean to add the real path that we use in our prov app deployments? We can add it to minimize the customization of this file but because this is a sample file I thought the value above is good enough... should we change it for the real file path... what do yo think?

michaelsauter: Yes I would use the actual path (basically what was hardcoded before). If I get this change correctly, then one can now change the path e.g. for local development, but one does not have to change it for a typical ODS installation, right?

stitakis: more or less... local environment is using the project application.properties which default to ```./build/logs```. Assuming this file is not just only an example but a kind of template I will add the actual path here.

stitakis: @michaelsauter changes done

clemensutschig:
@stitakis Please merge

clemensutschig:
@stitakis  - what are we doing with this now.. ?

clemensutschig:
@stitakis Ping?

stitakis:
no need to merge... this change is already in master in `ods-core-env.sample`. I'll close this.

PR #430, opened 2020-03-19
Add scripts to setup deployment target

michaelsauter:
create-target-project.sh:
This script can be used to create a new project, setup with serviceaccount
an image-puller secret so that the MRO can use it as a deploy target.
This script would typically be run by a cluster admin.

create-target-sa-secret.sh:
This script can be used to store the token of a serviceaccount with rights
to a deployment project in a secret, and share that secret with Jenkins.
This script would typically be run by a project admin.

@clemensutschig @oalyman @metmajer @gerardcl These are helper scripts to create new envs to deploy to. I'm targeting `master` here because - not sure if we should also put this in 2.x? Technically, nothing uses these scripts. They are just there as a reference implementation of how someone could setup these envs properly.

### DIFF
--- a/ocp-scripts/create-target-project.sh
+++ b/ocp-scripts/create-target-project.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+# This script can be used to create a new project, setup with serviceaccount
+# an image-puller secret so that the MRO can use it as a deploy target.
+# This script would typically be run by a cluster admin.
+set -eu
+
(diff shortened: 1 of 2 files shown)


michaelsauter:
I will merge them now so that the MRO documentation can reference the scripts - otherwise the part of the MRO doc describing the project setup is very fishy. We should still think about what to do with those scripts before they make it into the 3.0 release.

PR #431, opened 2020-03-20
Fix jenkins out of memory

### LINKED ISSUE #412: Jenkins runs out of memory and becomes unresponsive
**Describe the bug**
During our work on our internal ODS development environment, we increasingly see unresponsive Behavior with Jenkins. Analyzing the memory of the Java process has shown that we were too low on memory with our 2 GB settings: we ran out of heap space and had frequent garbage collector activity, which resulted in Jenkins becoming suspended frequently.

**Expected behavior**
Have a guess...

**Screenshots**

***OOM***
PastedGraphic-1

***Components failing randomly***
Screenshot 2020-02-14 at 16 23 30

***Non-availability due to frequent suspension***
Screenshot 2020-02-14 at 16 22 35

 - OpenDevStack 2

**Proposed Solution**
1) The following article suggests tuning parameters for Jenkins using the G1 GC: https://jenkins.io/blog/2016/11/21/gc-tuning/.
2) The following article says that `-Xmx` should be set to `Xms`. It also states that `-Xmx=2g` is the absolute minimum for Jenkins: https://support.cloudbees.com/hc/en-us/articles/204859670. 
3) The following article mentions a set of environment variables to be used to put the findings out of `1)` and `2)` into practice in OpenShift: https://docs.openshift.com/container-platform/3.9/using_images/other_images/jenkins.html.

Finally, here is a screenshot of a Deployment Config from our internal development cluster with an `-Xmx=4g`. Note that the applied resource limit's upper bound needs to be set to a higher value:

Screenshot 2020-02-14 at 17 50 27

hrcornejo:
Closes #412


hrcornejo:
@clemensutschig created #432 but that PR has more changes from 2.x

metmajer:
@michaelsauter @clemensutschig can this be merged?

create-projects/ocp-config/cd-jenkins/cd-jenkins-master.yml
```diff
@@ -19,11 +19,17 @@ parameters:
 - name: JENKINS_ENABLE_OAUTH
   value: "true"
 - name: JENKINS_MEMORY_LIMIT
-  value: 2Gi
+  value: 5Gi
```
michaelsauter: This won't fly. We run our cluster with a maximum 1:2 ratio for memory request/limit. And I assume many people do - as you should. If the ratio is to high, you're overcommitting the cluster too much, which is dangerous for memory.

Please change this to `4Gi` max. If you really need `5Gi` sometimes, we'd need to change the request to 2.5 or even 3 (the request should be close to what typical application usage is ...)

metmajer: Jenkins proposes to use all available memory for Heap space (we set this via `JENKINS_CONTAINER_HEAP_PERCENT` [here](https://github.com/opendevstack/ods-core/pull/431/files#diff-e52b91e71a932e8cd20b8dbd9e108f7cR31-R32)). When we set the `-Xmx` to 4GB [here](https://github.com/opendevstack/ods-core/pull/431/files#diff-e52b91e71a932e8cd20b8dbd9e108f7cR29-R30), we'll need to set the `JENKINS_MEMORY_LIMIT` to some value slightly higher (or reduce `JENKINS_CONTAINER_HEAP_PERCENT` to a lower value). I was not successful in setting `JENKINS_MEMORY_LIMIT` and `-Xmx` to the same value. In that case, Jenkins crashed. Will need to try this out.

michaelsauter: I see. What we definitely need to do though is to stay in the 1:2 ratio. I am wondering how you have been able to run this? If there is a cluster with a higher allowed ratio, that should be adjusted asap to avoid inconsistencies ...

metmajer: Got it. Will try out 4G and use a `JENKINS_CONTAINER_HEAP_PERCENT` of 0.95.

metmajer:
@hrcornejo Please reduce `JENKINS_MEMORY_LIMIT` to `4Gi` and leave everything else as-is. I have tested this set-up on OCTEST with `JENKINS_CONTAINER_HEAP_PERCENT=1` and this works.

PR #433, opened 2020-03-20
CNES scanner updates on ODS 2.x

clemensutschig:
update to cnes 3.1.0


clemensutschig:
@gerardcl - please update master as well then

PR #438, opened 2020-03-23
Fix Jenkins out of memory

### LINKED ISSUE #412: Jenkins runs out of memory and becomes unresponsive

hrcornejo:
Closes #412

### DIFF
--- a/create-projects/ocp-config/cd-jenkins/cd-jenkins-master.yml
+++ b/create-projects/ocp-config/cd-jenkins/cd-jenkins-master.yml
@@ -19,11 +19,17 @@ parameters:
 - name: JENKINS_ENABLE_OAUTH
   value: "true"
 - name: JENKINS_MEMORY_LIMIT
-  value: 2Gi
+  value: 4Gi
 - name: JENKINS_MEMORY_REQUEST
-  value: 1Gi
+  value: 2Gi
 - name: JENKINS_VOLUME_CAPACITY
   value: 5Gi
+- name: JENKINS_JAVA_GC_OPTS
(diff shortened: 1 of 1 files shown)


metmajer:
@michaelsauter @clemensutschig Thanks!

PR #448, opened 2020-03-25
Backport git scm plugin for 2.x branch

michaelsauter:
Required for
https://github.com/opendevstack/ods-jenkins-shared-library/issues/174.

FYI @borja44


michaelsauter:
@clemensutschig apparently this is mentioned in the install docs (I don't get why, but hey) so it was done manually in the test cluster. I figured we might as well put it into 2.x then given we want to roll out that plugin in 3.0 anyway.

PR #449, opened 2020-03-25
Remove operational scripts.

### LINKED ISSUE #447: Removal of non-essential operational scripts
I propose that we remove non-essential operational type scripts from ods-core. Having these operational scripts seem a bit odd to have in ods-core to me. 

In particular:

- scripts `check-unsecured-routes.sh` and `list-routes-status.sh`in [ods-core/ocp-scripts at ae327a3371](https://github.com/opendevstack/ods-core/tree/ae327a337146ba2ab8220d1e45d771af7525562e/ocp-scripts).
- the associated Jenkinsfile and Openshift config at [ods-core/check-ocp-secure-routes at ae327a3371](https://github.com/opendevstack/ods-core/tree/ae327a337146ba2ab8220d1e45d771af7525562e/check-ocp-secure-routes)

@michaelsauter, @gerardcl, @stefanlack, @tjaeschke , @kiwo : let me know if you have concerns or suggestions. Thanks!

henrjk:
Here is a PR of the proposed removals.

Closes #447


clemensutschig:
@henrjk  - why should they go out .. I think they are much needed .. @sino92  - thoughts?

henrjk:
@clemensutschig I was soliciting feedback in issue #447 in order to see whether removal makes sense to others as well. 

So this PR should only be applied if people in general agree with this. 

To me the operational scripts should not be in ods-core as they have a different purpose compared to the other stuff and can be customized for different organizations.

Instead of removing them there could also be a new place where these scripts could be moved to, perhaps to be called ods-contrib. This could be a place where one could contribute things which are not integral to ods. 

Another alternative is of course to do nothing and leave the scripts in place.

michaelsauter:
I like `ods-contrib` as an idea. Could also be something under the BIX-Digital organisation.

henrjk:
Indeed I found another reference at 
[update.adoc#L107](https://github.com/opendevstack/opendevstack.github.io/blame/8700d7e578f8cb535ecf93b8473ae15b670c28c6/docs/modules/administrator-guides/pages/update.adoc#L107)

Any thoughts whether this changes things?

Perhaps we could only remove list-routes-status.sh.

michaelsauter:
@henrjk I would add a note in that section of the guide saying that this feature has been removed in a later version so that admins can also skip this step.

I am still in favour of removing the whole thing - it is not CORE. You may choose to install it, or not. The check script should not depend on ODS, and should make sense in a vanilla OpenShift installation too.

henrjk:
I just added a commit for the removal in the Makefile. 

Let's wait a bit whether we have consensus on the removal or relocation to elsewhere and note that the update guide  will still need to be updated.

henrjk:
@clemensutschig, @sino92 what do you think?

clemensutschig:
@henrjk  - where are they going, new repo?

henrjk:
@clemensutschig Considering feedback my proposal is: 

- Go forward with this PR to remove the scripts from odd-core.
- Create an ods-contrib repo under the BIX-digital organization.
- Move the unsecured routes checking script to ods-contrib, but not the list_routes script.
- In the update guide add a note in that section of the guide saying that this feature has been removed in a later version so that admins can also skip this step. Also mention the ods-contrib repo.

clemensutschig:
ok - then go

PR #451, opened 2020-04-01
Ignore pushed refs other than "BRANCH" (e.g. "TAG") by default

### LINKED ISSUE #450: Webhook Proxy: Ignore pushed tags by default
Right now, any pushed Git refs create a new pipeline. Therefore, if tags get pushed, a new pipeline is created. This is most likely not needed. By default, it should be off, and possible to turn on via env var.

michaelsauter:
Closes #450.


michaelsauter:
@oalyman ping

Build failure is unrelated.

PR #453, opened 2020-04-03
Preserve immutable fields in Nexus/SonarQube config

michaelsauter:
E.g. if some config changed regarding PVCs, we do not want to destroy
the existsing PVC. Users may still choose to do that manually ...


michaelsauter:
@gerardcl ping

PR #455, opened 2020-04-03
docker build webhook-proxy in ci

henrjk:
Test simple docker build

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,10 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.(yml|yaml)]
+indent_size = 2
+
+
 [*.md]
 trim_trailing_whitespace = false ; trimming trailing whitespace may break Markdown
 
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -12,12 +12,6 @@ jobs:
         name: docker build webhook-proxy
         working-directory: jenkins/webhook-proxy
         run: |
-          echo GITHUB_ACTOR=$GITHUB_ACTOR
-          echo GITHUB_EVENT_NAME=$GITHUB_EVENT_NAME
-          echo GITHUB_SHA=$GITHUB_SHA
-          echo GITHUB_REF=$GITHUB_REF
-          echo GITHUB_EVENT_PATH=$GITHUB_EVENT_PATH
-          cat /github/workflow/event.json
           docker build -t webhook-proxy .
   webhook-proxy:
     name: Webhook Proxy tests
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -1,6 +1,18 @@
 name: Continous Integration Tests
 on: [push, pull_request]
 jobs:
+  build-jenkins-images:
+    name: Build Jenkins docker images
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: Checkout repository
+        uses: actions/checkout@v2.0.0
+      -
+        name: docker build webhook-proxy
+        run: |
+          cd jenkins/webhook-proxy
+          docker build -t webhook-proxy .
   webhook-proxy:
     name: Webhook Proxy tests
     runs-on: ubuntu-18.04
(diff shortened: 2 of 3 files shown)


clemensutschig:
@henrjk  / @oalyman  - how will we deal with internal self signed certificates? 
One of the reasons to build most of the images (internally) - is rooted in

1. https://github.com/opendevstack/ods-core/blob/master/jenkins/slave-base/Dockerfile.rhel7#L32-L45
1. https://github.com/opendevstack/ods-core/blob/master/jenkins/master/Dockerfile#L12-L19
1. https://github.com/opendevstack/ods-provisioning-app/blob/master/Jenkinsfile#L40-L41

this needs to be accounted for - because otherwise we'll not win much ...

henrjk:
@clemensutschig, yes this PR will not work as is because of the issue you mentioned, but also would not cover the rhel7 based images. I was wanting to experiment with github actions. I also hope to experiment some with UBI images when I find the time.

One thought was to have a base image which does not include the self-signed certificate customizations. Such base images could even be published to docker hub.

Our private builds could be based on these public images and do the self-signed certificate magic. 
However I have not looked into whether this would be feasible/make sense at all, so it is just a thought at the moment.
I mention also @michaelsauter as we had some brief discussion about image testing last week.

michaelsauter:
@clemensutschig We are already working on this. @oalyman Is fixing the certificates setup. The way it was done (baked into the image) is not the way to go, and one can do this at startup instead.

@henrjk If the webhook proxy build already works, why not make that a separate PR so that we can merge it already?

oalyman:
@clemensutschig @henrjk @michaelsauter I just created a PR (https://github.com/opendevstack/ods-provisioning-app/pull/406) that addresses the certificate authority topic for the prov-app to be able to publish an image to the public docker repo.

See this PR and the link below on the topic of Openshift and serving certificates:
https://docs.openshift.com/container-platform/3.11/dev_guide/secrets.html#service-serving-certificate-secrets.

> Other pods can trust cluster-created certificates (which are only signed for internal DNS names), by using the CA bundle in the /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt file that is automatically mounted in their pod

henrjk:
I wanted to note that in the most recent build checks associated with this PR jenkins-master was using the redhat base image for experimentation:

https://github.com/opendevstack/ods-core/blob/e139b85c4e7f1e49a0b6aba965ead9258b5abc40/jenkins/master/Dockerfile#L1

While the base image can be pulled from redhat without authentication, it fails when it needs entitlement during install of openssl and gnutls-utils as shown below:

```txt
...
Step 9/20 : RUN yum -y install openssl gnutls-utils     && $JAVA_HOME/bin/keytool -storepasswd -new mysecretpassword -keystore $JAVA_HOME/lib/security/cacerts -storepass changeit     && yum clean all      && rm -rf /var/cache/yum/*
 ---> Running in 2787858d88a3
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not registered with an entitlement server. You can use subscription-manager to register.
[91mThere are no enabled repos.
 Run "yum repolist all" to see the repos you have.
 To enable Red Hat Subscription Management repositories:
     subscription-manager repos --enable <repo>
 To enable custom repositories:
     yum-config-manager --enable <repo>
```

oalyman:
@henrjk why not using the UBI?

henrjk:
After reading docs I believe there is no ubi for openshift v3.11 so that we would have to cook this ourselves. However I did not yet double check this by engaging with  https://github.com/openshift/jenkins yet.

oalyman:
@henrjk Ah sorry. Just realised that you are talking about the jenkins base image.
Yes so far I am also not aware of a jenkins base image that uses UBI.

oalyman:
@henrjk I just read that with Openshift 4 the jenkins image will be based on UBI (https://github.com/openshift/jenkins)

> For OpenShift v4, the operating system choice is reduced. All OpenShift v4 images (including the ones from this repository) are based off of the "Universal Based Image" or "UBI".

henrjk:
Yes. One aspect which I have not drilled down on is what makes the v4 jenkins images openshift v4 specific. 
My read of (https://github.com/openshift/jenkins) is that the v4 images are currently RHEL7 based. @oalyman if want to experiment or drill down keep me updated.

michaelsauter:
Replaced by #592.

PR #456, opened 2020-04-06
ci build jenkins images: webhook-proxy

### LINKED ISSUE #157: Build the webhook proxy image in Travis
Ideally we could build the image to see if that works. That would protect us against bugs like #151.

henrjk:
See https://github.com/opendevstack/ods-core/pull/455#issuecomment-609605649

However perhaps it would also make sense to have this on docker hub. For this I see a need to at least:

1. Figure out how to tag this?
2. Establish a docker Team or Organization for BI or BIX (see https://docs.docker.com/docker-hub/orgs/). 

Closes #157.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,9 @@ trim_trailing_whitespace = true
 [*.md]
 trim_trailing_whitespace = false ; trimming trailing whitespace may break Markdown
 
+[*.(yml|yaml)]
+indent_size = 2
+
 [Makefile]
 tab_width = 4
 indent_style = tab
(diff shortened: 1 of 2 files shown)


clemensutschig:
> hat it should be consistent - as in, either all Jenkins images are built on Github, or none. With the Jenkins Master, it isn't easy (right now) as that has a file in it which depends on local config.

why NOT on github in our org - this is now fully supported (tab pacakges)

michaelsauter:
@clemensutschig Good point about GitHub being able to store images. I moved this to another issue: https://github.com/opendevstack/ods-core/issues/490

However, my comment was about pushing the image publicly at all. I moved that to a separate ticket as well: https://github.com/opendevstack/ods-core/issues/491

This PR is just about building an image, which I think is fine as-is.

PR #461, opened 2020-04-07
Use latest version (1.0.44) of openshift-sync plugin

### LINKED ISSUE #441: Figure out highest possible openshift-sync plugin version
We have effectively fixed the version of the [openshift-sync-plugin](https://github.com/jenkinsci/openshift-sync-plugin) to pre-1.0.33 since #440. A recent update by RedHat made Jenkins ship with 1.0.43 out of the box (which would then be overwritten with our patched JAR if Jenkins was deployed again).

1.0.42 has a bug which prevents source secrets from being synced, and 1.0.43 has a bug which sets the "refspec" field which prevents new commits from being picked up by Jenkins.

1.0.44 is the newest version, and we do not know if it works (but from the commit history it looks largely identical to 1.0.43, which suggests that it is broken, too).

We should figure out which one is the highest, bugfree version, and we should follow up with RedHat to get the "refspec bug" fixed. 

FYI @clemensutschig @gerardcl @oalyman

michaelsauter:
Closes #441.

I deleted the PVC of Jenkins and created new pipelines - `refspec` is not set as expected. That means 1.0.44 is correct.

### DIFF
--- a/jenkins/master/plugins.txt
+++ b/jenkins/master/plugins.txt
@@ -8,4 +8,4 @@ workflow-cps-global-lib:2.13
 junit:1.28
 blueocean:1.18.0
 workflow-durable-task-step:2.28
-openshift-sync:1.0.34
+openshift-sync:1.0.44


gerardcl:
@michaelsauter should we add this as an update note for existing projects with older ODS versions? the users will require recreating the existing PVC, right?

michaelsauter:
@gerardcl Yes, or they need to update via the UI, which is probably what we should suggest instead.

gerardcl:
proposal for update guides: shall we create a new label "update guide", or similar name, so to remember/search/list on the go what should be mentioned in the guides when new releases come?

michaelsauter:
@gerardcl Good idea, added!

PR #471, opened 2020-04-09
Unify starting and following builds

michaelsauter:
* Fixes a few things that are broken in the test setup right now
* Unifies starting/following/checking of builds
* Does not run Tailor non-interactively by default anymore (except in tests)

### DIFF
--- a/ocp-scripts/start-and-follow-build.sh
+++ b/ocp-scripts/start-and-follow-build.sh
@@ -51,4 +51,5 @@ for i in 1 2 3; do
   fi
   sleep 3
 done
+echo "Please, check build ${BUILD_ID} since it seems failed/not completed."
 exit 1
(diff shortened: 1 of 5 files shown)


ocp-scripts/start-and-follow-build.sh
```diff
@@ -0,0 +1,54 @@
...
+
+for i in 1 2 3; do
+  BUILD_STATUS=$(oc -n ${NAMESPACE} get build ${BUILD_ID} -o jsonpath='{.status.phase}')
+  if [ "${BUILD_STATUS}" == "Complete" ]; then
+    echo "Build ${BUILD_ID} is complete."
+    exit 0
+  fi
+  sleep 3
+done
+exit 1
```
gerardcl, line 54:
```suggestion
echo "Please, check build ${BUILD_ID} since it seems failed/not completed."
exit 1
```

PR #472, opened 2020-04-09
Add script to initialise, update and sync repos

michaelsauter:
* If you have no repos setup, it allows you to clone them.
* If you have repos already, it allows to update them.
* It allows to sync branches between GitHub and BitBucket.

We can use this script as a "getting started" like this:
```
mkdir -p opendevstack
cd opendevstack
curl -LO https://raw.githubusercontent.com/opendevstack/ods-core/master/ocp-setup/repos.sh
chmod +x repos.sh
./repos.sh --init
```

@gerardcl I believe this to be a huge simplification when doing the installation. In the past, that was always a way lengthier process than it should have been, with lots of manual errors.

### DIFF
--- a/ods-setup/setup-repos.sh
+++ b/ods-setup/setup-repos.sh
@@ -1,0 +1,172 @@
+
+#!/usr/bin/env bash
+set -ue
+
+echo_done(){
+  echo "\033[92mDONE\033[39m: $1"
+}
+
+echo_error(){
+  echo "\033[31mERROR\033[39m: $1"
+}
+
+echo_info(){
+  echo "\033[94mINFO\033[39m: $1"
+}
+
+CONFIRM=
+BITBUCKET_URL=
+GIT_REF=
+PUSH=
+
+function usage {
+  printf "Initialise, update and sync OpenDevStack repositories.\n\n"
+  printf "This script will ask interactively for parameters by default.\n"
+  printf "However, you can also pass them directly. Usage:\n\n"
+  printf "\t-h|--help\t\tPrint usage\n"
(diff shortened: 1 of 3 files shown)


ods-setup/setup-repos.sh
```diff
@@ -0,0 +1,172 @@
+
+#!/usr/bin/env bash
+set -ue
+
+echo_done(){
+  echo "\033[92mDONE\033[39m: $1"
```
sino92, line 6: Any reason why you omitted `-e`?

ods-setup/setup-repos.sh
```diff
@@ -0,0 +1,172 @@
...
+fi
+if [ -z ${GIT_REF} ]; then
+  echo_error "Git ref cannot be empty.";
+  exit 1;
+else
+  echo_info "Repos will be checked out @ ods/${GIT_REF}."
+fi
+
+if [ -z ${PUSH} ]; then
+  read -e -p "Do you want to push ods/${GIT_REF} to your BitBucket server? [y/n] " input
```
sino92, line 93: Repeated typo BitBucket -> Bitbucket

sino92:
The script saves quite some time, as opposed to jumping around different projects syncing up each individually. :) Great stuff! 
One of the aims of this was to make BI adapt ODS release updates easier, right? Aside from that, its not yet adapted to support other organizations' contributions. It seems the script assumes the given bitbucket account already contains a fork of ODS. As for independent contributors working with Bitbucket, they will need to fork each repo individually if I understood correctly.

PR #489, opened 2020-04-22
add provisioning-app configuration to run in "cd" namespace

oalyman:
This enables one to run a prebuild image (e.g. opendevstackorg/ods-provisioning-app:latest) in the cd namespace by using `tailor apply --namespace cd`


configuration-sample/ods-core.env.sample
```diff
@@ -161,3 +161,54 @@ STORAGE_CLASS_DATA=""
...
+
+# which Docker image of the prov app to run
+PROV_APP_IMAGE=opendevstackorg/ods-provisioning-app:latest
+
+# which image-pull-policy to use 
+# see https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#image-pull-policy
+PROV_APP_IMAGE_PULL_POLICY=Always
+
+# confluence url - including protocol
+PROV_APP_CONFLUENCE_URI=http://192.168.56.31:8090
```
michaelsauter, line 180: I would change this, and the URI's below, to just `CONFLUENCE_URL`, `BITBUCKET_URL`. That way users do not have to specify it twice.

provisioning-app/ocp-config/Tailorfile
```diff
@@ -0,0 +1,8 @@
+namespace cd
+selector app=prov-app
+param-file ../../../ods-configuration/ods-core.env
+ignore-path bc:/spec/output/to/name
+ignore-path bc:/spec/output/imageLabels
```
michaelsauter, line 5: As there is no `BuildConfig` resource, we should be able to remove those ignored paths.

provisioning-app/ocp-config/Tailorfile
```diff
@@ -0,0 +1,8 @@
...
+ignore-unknown-parameters true
+
+bc,cm,dc,pvc,route,secret,svc
```
michaelsauter, line 8: `bc` should not be needed?

provisioning-app/ocp-config/dc.yml
```diff
@@ -0,0 +1,69 @@
...
+      resources: {}
+      type: Recreate
+    template:
+      metadata:
+        labels:
+          app: prov-app
+      spec:
+        containers:
+        - image: ${PROV_APP_IMAGE}
+          imagePullPolicy: ${PROV_APP_IMAGE_PULL_POLICY}
```
michaelsauter, line 32: This is not defined in the `parameters` list. That should break, no?

provisioning-app/ocp-config/pvc.yml
```diff
@@ -0,0 +1,24 @@
...
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 10Mi
+parameters:
+- name: STORAGE_PROVISIONER
+- name: STORAGE_CLASS_DATA
+- name: STORAGE_CLASS_BACKUP
```
michaelsauter, line 24: For consistency, can we mark those required?

configuration-sample/ods-core.env.sample
```diff
@@ -161,3 +161,54 @@ STORAGE_CLASS_DATA=""
 
 # Storage class for backup, for AWS this should be "gp2-encrypted"
 STORAGE_CLASS_BACKUP=""
+
+
+####################
+# Provisioning App #
+####################
+
+PROV_APP_ENVIRONMENT=cd
```
michaelsauter, line 170: Seeing the later use of this, it is the same as the automatic parameter `TAILOR_NAMESPACE`. Can we remove the sample param here, and make use of `TAILOR_NAMESPACE` in the templates?

provisioning-app/ocp-config/secret.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Template
+objects:
+- apiVersion: v1
```
michaelsauter, line 4: Is this secret being used at all? I suspect it was only there to help build the image (to clone the repo).

clemensutschig:
can you hold this back until we have decided opendevstack/ods-core#493

stitakis:
Jan made me aware yesterday that even if ProvApp is deployed in CD namespace, it will still depend on the jenkins and webhook deployed in prov-cd namespaces. As next step, we will need to additionally deploy those components into cd namespace to have all in one place.

michaelsauter:
I'd actually vote for merging this once feedback is addressed, even if we have not decided #493 yet. After all, we are just merging into `master` and can adjust the target later. The only thing right now that hardcodes `cd` is in the `Tailorfile`. If you pass the namespaceto Tailor (`tailor apply -n foo`) you will get a prov -app in the `foo` namespace :)

PR #514, opened 2020-04-30
install jenkins and prov-app in ods namespace

### LINKED ISSUE #502: Move Jenkins ocp-config out of create-projects
In preparation to have a Jenkins running in the global ODS namespace, it would be good to move the `ocp-config` for installing Jenkins to another folder, maybe closer to `jenkins`. There we have a config that applies to "build" resources, and the config that is in `create-projects` right now could be referred to "deploy" resources (as is already the naming convention used in the `Makefile`).

FYI @oalyman

oalyman:
Resolves #502


Makefile
```diff
@@ -56,6 +61,11 @@ start-jenkins-build-webhook-proxy:
 	ocp-scripts/start-and-follow-build.sh --build-config jenkins-webhook-proxy
 .PHONY: start-jenkins-build-webhook-proxy
 
+# PROVISIONING APP
+## Install the gloabl provision app for the cluster
+install-provisioning-app:
+	cd provisioning-app/ocp-config && tailor apply --namespace ods --selector app=prov-app
```
michaelsauter, line 67: I think it would be safer if we move the selector into the `Tailorfile` - to avoid accidental damage.

michaelsauter:
Doing my suggested change would also automatically fix the tests which are broken due to the changed file structure.

clemensutschig:
Super cool ... Re Provision App..  Please Change the selector .. it changed with my Merge.  And add the doc gen Service as well.

PR #515, opened 2020-04-30
add doc gen service and new selectors

clemensutschig:
add document generation service .. and use consistent selectors now that match ods-quicsktarters :)

### DIFF
--- a/ods-document-generation-svc/openshift/Tailorfile
+++ b/ods-document-generation-svc/openshift/Tailorfile
@@ -1,0 +1,8 @@
+namespace cd
+ignore-unknown-parameters true
+
+is
+
+selector app=ods-doc-gen-svc
+
+preserve bc:/spec/output/to/name,bc:/spec/output/imageLabels
--- a/ods-document-generation-svc/openshift/Tailorfile
+++ b/ods-document-generation-svc/openshift/Tailorfile
@@ -1,8 +1,6 @@
 namespace cd
 ignore-unknown-parameters true
 
-is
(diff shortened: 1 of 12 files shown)


ods-document-generation-svc/openshift/Tailorfile
```diff
@@ -0,0 +1,8 @@
+namespace cd
+ignore-unknown-parameters true
+
+is
```
michaelsauter, line 4: I think this is missing other resource types, right?

ods-document-generation-svc/openshift/Tailorfile
```diff
@@ -0,0 +1,8 @@
+namespace cd
+ignore-unknown-parameters true
+
+is
+
+selector app=ods-doc-gen-svc
+
+preserve bc:/spec/output/to/name,bc:/spec/output/imageLabels
```
michaelsauter, line 8: There shouldn't be any `BuildConfig`, so this line can probably be removed.

ods-document-generation-svc/openshift/is.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: 'ods-document-generation-service'
+  app: 'ods-doc-gen-svc'
```
michaelsauter, line 5: Not sure if elsewhere we have labels on the template resource itself. If so, we can just remove them, as we process the template locally, it does not need labels.

PR #522, opened 2020-05-04
Use image for contribution flow

michaelsauter:
That way we can get rid of the plantuml dependency to make it easier to
render documentation locally, see
https://github.com/opendevstack/ods-documentation/issues/46.


michaelsauter:
Followed by https://github.com/opendevstack/ods-documentation/pull/51.

PR #528, opened 2020-05-05
Change default central namespace from "cd" to "ods"

### LINKED ISSUE #493: Move opendevstack from current "global" CD OCP namespace to opendevstack/ods namespace
I gave this some thought - and the longer I think about it  - the more I believe we should do this ..

1. it gives clarity
1. for a user during upgrade there is no difference between changing an `image ref` in the a components' `jenkinsFile` - or the entire path of the image used there (from `cd/....:2.x` to `opendevstack/....3.x`) and similar for jenkins / webhook / docgen dc 
1. We could flip all project specific service account' pull rights to one rolebinding `image puller system:authenticated` so no impact for a user on that front
1. In case we ship published images, we have to create image streams pointing to those published images anyway in the global namespace, no matter if it's in the current 'CD' or the new `opendevstack` one.. 

On our side the change is massive (but the outcome is really nice, as it could be configurable via ods-configuration env properties) .. so people could actually have multiple namespaces, including `-dev/test` ones..

Changes needed:
1. quickstarters (jenkins slave references / master imagestream)
1. ods-core (all master image streams, scripts, role bindings for pull rights)
1. prov cd (jenkins image)
1. mro / doc gen service (imagestreams, jenkins slave images)

@michaelsauter @metmajer @gerardcl @sino92  - thoughts?

michaelsauter:
- Allows to overwrite the default namespace
- Updates scripts to take namespace parameter where needed
- Fixes a few smaller bugs in the Makefile
- Adds install-provisioning-app Makefile target for consistency
- Changes folder names to ocp-config for consistency
- Adds installation of doc gen service

Closes #493.


ocp-scripts/import-project.sh
```diff
@@ -506,12 +506,6 @@ do
 			echo
 		fi
 
-		# add pull rights against ocp OD dedicated - for NON CD projects - otherwise we even pull the jenkins image and that we dont want
-		if [[ $ocp_proj_namespace_suffix == "cd" ]]; then
-			echo "--  not creating pull secret to OCP, this is CD - just adding image pull rights"
-			oc policy add-role-to-user system:image-puller system:serviceaccount:${project_name}-cd:jenkins -n cd
-		fi
```
michaelsauter, line 513: FYI This is not required any longer as everyone is allowed to pull from central namespaces.

Issue #468, opened 2020-04-08
Test setup contains code that is not present in Makefile


michaelsauter:
E.g. https://github.com/opendevstack/ods-core/blob/master/tests/scripts/recreate-test-infrastructure.sh#L65.

This command is required for ODS to work, but it is only present in the script used by the test. 

We should change this (and any other place doing similar things) so that:

* The `Makefile` targets execute all commands required to install ODS
* The tests use the `Makefile` targets

FYI @hugowschneider @georgfedermann @oalyman

michaelsauter:
Fixed in #600.

PR #529, opened 2020-05-05
Assign self-provisioner role to Jenkins serviceaccount

### LINKED ISSUE #380: Solve self-provisioner permissions mystery
This is a follow-up from https://github.com/opendevstack/ods-project-quickstarters/issues/314.

ODS has the following requirements:

1. The `jenkins` serviceaccount in the `prov-cd` project needs to be able to create new projects, as it is responsible to create the `foo-dev`, `foo-test`, `foo-cd` namespaces.
2. The `jenkins` serviceaccount in each `*-cd` project needs to be able to create new projects when the auto-clone feature is used.

Observing the situation in different clusters:

* On a RedHat Dedicated cluster, both use-cases work, without granting `self-provisioner` explicitly anywhere in ODS. The rights must come from somewhere else. Potentially https://access.redhat.com/solutions/2988521 has been applied? I will follow this up to see if it is the case. Still, it would not explain how projects created through `prov-cd:jenkins` can make use of the auto-clone feature (2). There must be some other secret sauce that I do not know yet ...
* In the tests running in a temporary cluster using `oc cluster up`, we have now realised that the rights to do this are actually applied outside of ODS code - they are applied just in the tests. See https://github.com/opendevstack/ods-core/blob/f96c2ad2b25ac808f33d6b6b6f53dbbf5c7d8043/tests/scripts/recreate-test-infrastructure.sh#L12-L14 and https://github.com/opendevstack/ods-core/blob/f96c2ad2b25ac808f33d6b6b6f53dbbf5c7d8043/tests/scripts/recreate-test-infrastructure.sh#L65 The tests do not cover the auto-clone feature yet so it is unclear if auto cloning would work given those settings (in theory they should not ...). As a follow-up, the tests should NOT apply those changes in test-code, the commands need to move to ODS setup-code.
* In an onprem cluster updated from ODS 1.2 to 2.0, both use cases do not work without granting rights like it is done in the tests.

FYI @clemensutschig @rattermeyer @hugowschneider @gerardcl @henrjk @stitakis @stefanlack 

Please add any piece of information you have that I have not yet written down. I am slowly starting to see the whole picture and I do not want to miss something.

michaelsauter:
Closes #380.

Solves part of #468.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ prepare-config:
 # ODS SETUP
 ## Setup central "ods" project.
 install-ods-project:
-	cd ods-setup && ./setup-ods-project.sh --namespace ${NAMESPACE}
+	cd ods-setup && ./setup-ods-project.sh --namespace ${NAMESPACE} --reveal-secrets
 
 
 # JENKINS
--- a/ods-setup/ocp-config/cd-user/Tailorfile
+++ b/ods-setup/ocp-config/cd-user/Tailorfile
@@ -1,5 +1,5 @@
 param-file  ../../../../ods-configuration/ods-core.env
-
+selector template=cd-user-template
 ignore-unknown-parameters true
 
 secret
--- a/ods-setup/ocp-config/cd-user/secret.yml
+++ b/ods-setup/ocp-config/cd-user/secret.yml
@@ -1,20 +1,20 @@
 apiVersion: v1
 kind: Template
-labels:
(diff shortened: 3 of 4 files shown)


michaelsauter:
@clemensutschig Added hints.

michaelsauter:
FYI @stitakis @georgfedermann

PR #534, opened 2020-05-05
Fix spots with hardcoded cd namespace

michaelsauter:


sonarqube/ocp-config/sonarqube.yml
```diff
@@ -2,6 +2,74 @@ apiVersion: v1
 kind: Template
 metadata:
   name: sonarqube
+parameters:
```
michaelsauter, line 5: Just moved these from the bottom of the template to the top for consistency (and added the `ODS_NAMESPACE` param)

PR #538, opened 2020-05-06
Make Bitbucket project configurable

### LINKED ISSUE #347: Configure Bitbucket "opendevstack" project name
Some users might want to have multiple ODS installations (as in, multiple OpenShift clusters) using the same Bitbucket instance, but separate sources, e.g. because one installation serves as a playground / to test updates etc.

Allow to configure the `opendevstack` part in URLs etc.

### LINKED ISSUE #518: Fetch shared lib from github per default instead of requiring "local" clone on bitbucket
With the new ods default setup - I am wondering if we should not fetch the shared lib directly from github per default?!

This would allow us to remove yet another "install step" ... so instead of 
https://github.com/opendevstack/ods-core/blob/master/jenkins/master/configuration/init.groovy.d/ods-jenkins-shared-library.groovy#L17

we would seed the github url :)

@michaelsauter?

michaelsauter:
Closes #347.

Closed #518.


jenkins/ocp-config/build/bc.yml
```diff
@@ -2,6 +2,35 @@ apiVersion: v1
...
+- name: APP_DNS_PORT
+  description: OpenShift application base dns port - used for grabbing the root ca and adding into the agent
+- name: JENKINS_AGENT_BASE_SNYK_DISTRIBUTION_URL
+  description: optional uri that points to the snyk binary distribution (i.e. https://github.com/snyk/snyk/releases/download/v1.180.1/snyk-linux)
+- name: TARGET_HOSTS
+  description: if you have multiple DNS with different root ca, add them here seperated by ';'. They are added to jenkins master
+- name: SONAR_EDITION
+  description: SonarQube edition. One of "community", "developer", "enterprise" or "datacenter".
+- name: SONAR_VERSION
+  description: SonarQube version, e.g. "7.9"
```
michaelsauter, line 33: FYI I just moved this parameters section to the top for consistency.

PR #540, opened 2020-05-06
Check scripts with shellcheck

michaelsauter:
See https://www.shellcheck.net/.


michaelsauter:
@henrjk I think we should. But most of `infrastructure-setup` will probably just go away because it is outdated, and for the other folders I think we should do it bit-by-bit ... when we get time. I just wanted to start somewhere :)

PR #547, opened 2020-05-12
add import of ods doc & prov app

### LINKED ISSUE #545: Makefile should tag dockerhub images (scheduled) into respective image streams
With ods 3 - we will provide imagestreams and docker images at dockerhub for provision app and doc gen service. 
Yes for now people need to manually import those images into the right imagestreams - namely:

1. https://github.com/opendevstack/ods-core/blob/master/ods-document-generation-svc/ocp-config/is.yml
1. https://github.com/opendevstack/ods-core/blob/master/ods-provisioning-app/ocp-config/is.yml

The reason why want to have imaestreams, is 2 fold
1. people can switch between the tags in a controlled manner, and 
1. in case the want to develop - they can just tag the "locally" built image into the main imagestream and refer to it from the dc

Scheduled imports - as in https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#image-pull-policy 
``` 
apiVersion: v1
kind: ImageStream
metadata:
  name: ruby
spec:
  tags:
  - from:
      kind: DockerImage
      name: openshift/ruby-20-centos7
    name: latest
    importPolicy:
      scheduled: true
``` 
or in the setup / make file ...
```
$ oc tag docker.io/opendevstackorg/ods-provisioning-app odsst-dev/ods-provisioning-app:latest --scheduled
```
I strongly vouch that both prov app points to the imagestream and NOT direclty from the DC to a version at dockerhub.

clemensutschig:
closes #545

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ install-doc-gen: apply-doc-gen-build
 ## Update OpenShift resources related to the Document Generation image.
 apply-doc-gen-build:
 	cd ods-doc-gen-svc/ocp-config && tailor apply --namespace ${NAMESPACE}
-	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-document-generation-svc targetimagestream ods-doc-gen-svc
+	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-document-generation-svc --targetimagestream ods-doc-gen-svc
 .PHONY: apply-doc-gen-build
 
 
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,9 @@ install-provisioning-app: apply-provisioning-app-deploy
 
 ## Update OpenShift resources related to the Provisioning App service.
 apply-provisioning-app-deploy:
-	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE}
+	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE} is
+	../../ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-provisioning-app
+	tailor apply --namespace ${NAMESPACE} --exclude is
 .PHONY: apply-provisioning-app-deploy
 
 
@@ -87,6 +89,7 @@ install-doc-gen: apply-doc-gen-build
 ## Update OpenShift resources related to the Document Generation image.
 apply-doc-gen-build:
 	cd ods-doc-gen-svc/ocp-config && tailor apply --namespace ${NAMESPACE}
+	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-document-generation-svc targetimagestream ods-doc-gen-svc
 .PHONY: apply-doc-gen-build
 
 
--- a/Makefile
+++ b/Makefile
@@ -76,8 +76,8 @@ install-provisioning-app: apply-provisioning-app-deploy
 ## Update OpenShift resources related to the Provisioning App service.
 apply-provisioning-app-deploy:
 	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE} is
-	../../ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-provisioning-app
-	tailor apply --namespace ${NAMESPACE} --exclude is
+	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --image ods-provisioning-app
+	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE} --exclude is
 .PHONY: apply-provisioning-app-deploy
 
 
@@ -89,7 +89,7 @@ install-doc-gen: apply-doc-gen-build
 ## Update OpenShift resources related to the Document Generation image.
 apply-doc-gen-build:
 	cd ods-doc-gen-svc/ocp-config && tailor apply --namespace ${NAMESPACE}
-	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-document-generation-svc --targetimagestream ods-doc-gen-svc
+	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --image ods-document-generation-svc --target-stream ods-doc-gen-svc
 .PHONY: apply-doc-gen-build
 
 
--- a/ocp-scripts/import-image-from-dockerhub.sh
+++ b/ocp-scripts/import-image-from-dockerhub.sh
@@ -21,10 +21,10 @@ echo_info(){
 }
 
 ODS_AT_DOCKERHUB=docker.io/opendevstackorg
-IMAGE_NAME=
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -75,7 +75,9 @@ install-provisioning-app: apply-provisioning-app-deploy
 
 ## Update OpenShift resources related to the Provisioning App service.
 apply-provisioning-app-deploy:
-	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE}
+	cd ods-provisioning-app/ocp-config && tailor apply --namespace ${NAMESPACE} is
+	../../ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-provisioning-app
```
michaelsauter, line 79: Path looks wrong - I think each line in the `Makefile` should be executed in a new shell, so this should be relative to the `Makefile` itself (as you have it in `apply-doc-gen-build`).

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+    -h|--help) usage; exit 0;;
+
+    -on|--namespace) NAMESPACE="$2"; shift;;
+    -on=*|--namespace=*) NAMESPACE="${1#*=}";;
+
+    -in|--imagename) IMAGE_NAME="$2"; shift;;
+    -in=*|--imagename=*) IMAGE_NAME="${1#*=}";;
+
+    -iv|--imageversion) IMAGE_VERSION="$2"; shift;;
+    -iv=*|--imageversion=*) IMAGE_VERSION="${1#*=}";;
```
michaelsauter, line 55: Did you pick `IMAGE_VERSION` intentionally over `IMAGE_TAG`? I think the typical container parlance would be "tag".

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+    *) echo_error "Unknown parameter passed: $1"; exit 1;;
+esac; shift; done
+
+if ! which jq >/dev/null; then
+    echo_warn "Binary 'jq' (https://stedolan.github.io/jq/) is not in your PATH. This will make the script less comfortable to use."
+    read -r -e -p "Continue anyway? [y/n] " input
+    if [ "${input:-""}" != "y" ]; then
+        exit 1
+    fi
+fi
```
michaelsauter, line 69: I guess this is leftover, right? No need for `jq` here ...

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+        echo_info "Configuration located"
+        odsprojectname=$(grep NAMESPACE "${ODS_CORE_DIR}/../ods-configuration/ods-core.env" | cut -d "=" -f 2-)
+    fi
+    read -r -e -p "Enter ods central namespace [${odsprojectname}]: " input
+    if [[ -z "${input}" || $(echo -n $input | wc -m) == 0 ]]; then
+        NAMESPACE=${odsprojectname}
+    else
+        NAMESPACE=${input:-""}
+    fi
+fi
```
michaelsauter, line 83: Hmm, I think we can delete this whole thing. You default `NAMESPACE` to `ods` at the top, so it should never be empty (apart from the user setting it to `""`, but I guess we should just fail in that case).

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+    else
+        IMAGE_NAME=${input:-""}
+    fi
+fi
+
+if [ -z "${TARGET_IMAGE_STREAM}" ]; then
+    read -r -e -p "Enter target image stream name: " input
+    if [ -z "${input}" ]; then
+        TARGET_IMAGE_STREAM=$IMAGE_NAME
+        echo "Target not set, defaulting to source: $IMAGE_NAME"
```
michaelsauter, line 104: I think it would be nice to display the default in the prompt, such as: `Enter target image stream name [foobar]: `. When the user hits enter, `foobar` is assigned.

michaelsauter: Ah -  I see you already used this pattern in line 121 :+1:

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+    fi
+fi
+
+TAG_SOURCE=${ODS_AT_DOCKERHUB}/${IMAGE_NAME}:${IMAGE_VERSION}
+TAG_TARGET=${NAMESPACE}/${TARGET_IMAGE_STREAM}:${IMAGE_VERSION}
+
+echo_info "importing remote image ${TAG_SOURCE} into local ${TAG_TARGET}"
+
+# tag directly - this should fail after a couple seconds, so we can verify
+oc tag ${TAG_SOURCE} ${TAG_TARGET} --scheduled
```
michaelsauter, line 135: I don't get the comment above. Does that mean that this command always succeeds, even if the "import" fails?

ocp-scripts/import-image-from-dockerhub.sh
```diff
@@ -0,0 +1,144 @@
...
+    -on=*|--namespace=*) NAMESPACE="${1#*=}";;
+
+    -in|--imagename) IMAGE_NAME="$2"; shift;;
+    -in=*|--imagename=*) IMAGE_NAME="${1#*=}";;
+
+    -iv|--imageversion) IMAGE_VERSION="$2"; shift;;
+    -iv=*|--imageversion=*) IMAGE_VERSION="${1#*=}";;
+
+    -it|--targetimagestream) TARGET_IMAGE_STREAM="$2"; shift;;
+    -it=*|--targetimagestream=*) TARGET_IMAGE_STREAM="${1#*=}";;
```
michaelsauter, line 58: Minor nitpick: Elsewhere we use hyphens for multi-word-params, so to align this would be `--target-image-stream`. Also, not a huge fan of multi-letter params with single hyphen (like `-it`) because often it is possible to combine multiple single-hypen-params (e.g. Docker allows you to do `docker exec -it foo bar`, which is equal to `docker exec -i -t foo bar`).

Maybe `s|stream`, `t|tag`, `n|namespace` ,`i|image` would work too?

Makefile
```diff
@@ -87,6 +89,7 @@ install-doc-gen: apply-doc-gen-build
 ## Update OpenShift resources related to the Document Generation image.
 apply-doc-gen-build:
 	cd ods-doc-gen-svc/ocp-config && tailor apply --namespace ${NAMESPACE}
+	ocp-scripts/import-image-from-dockerhub.sh --namespace ${NAMESPACE} --imagename ods-document-generation-svc --targetimagestream ods-doc-gen-svc
```
michaelsauter, line 92: Should we just rename the image stream? Since we move to `ods` namespace anyway ...

michaelsauter:
I like it 👍

ocp-scripts/import-image-from-dockerhub.sh
clemensutschig: take a look into the comment now - tag is async in case you import from a foreign source ... so we need to wait, and then try to get the istag

clemensutschig:
@michaelsauter  - thanks for the comments, I like the change requests a lot, and the feedback2

PR #563, opened 2020-05-18
base prov-app deployment on image stream

oalyman:


ods-provisioning-app/ocp-config/dc.yml
```diff
@@ -68,4 +68,13 @@ objects:
           name: volume-ikfik
     test: false
     triggers:
+    - imageChangeParams:
+        automatic: true
+        containerNames:
+        - prov-app
+        from:
+          kind: ImageStreamTag
+          name: prov-app:latest
```
michaelsauter, line 77: This would need to be `${PROV_APP_IMAGE}`, right?

PR #566, opened 2020-05-19
edit role is not a cluster role

michaelsauter:
@gerardcl Do you have any idea why it was like this? I am confused, it should not be a `ClusterRole`.

FYI @georgfedermann you might also run into this ...

`userNames` and `groupNames` is deprecated in favour of `subjects` so not needed.


gerardcl:
I think this is a legacy configuration from the very very beginning, which I even don't think this was required but kept because of manual/test installations
could be that in previous OCP versions that was required? and/or the intention maybe was to be able to create projects for the autocloning? I definitely don't know :(

PR #569, opened 2020-05-20
change service-ca import to runtime for sonarcube and jenkins

oalyman:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -15,6 +27,14 @@ jobs:
...
         run: |
           shellcheck sonarqube/*.sh
+      -
+        name: Build image
+        working-directory: sonarqube
+        run: |
+          docker build \
+          --build-arg sonarVersion=${{ matrix.version }} \
+          --build-arg sonarDistributionUrl="https://binaries.sonarsource.com/Distribution/sonarqube/sonarqube-${{ matrix.version }}.zip" \
+          -t sonarcube .
```
michaelsauter, line 37: `test.sh` already builds the image, so this should not be necessary?

jenkins/master/Dockerfile
```diff
@@ -36,6 +32,8 @@ RUN mkdir -p /etc/opendevstack \
     && echo "  \"sonarqubeVersion\": \"$SONAR_VERSION\"" >> /etc/opendevstack/config.json \
     && echo "}" >> /etc/opendevstack/config.json
 
+RUN chmod g+w $JAVA_HOME/lib/security/cacerts
```
michaelsauter, line 35: Were you able to verify that this works both RedHat and CentOS based? It looks right to me, but would be good to verify.

.github/workflows/continuous-integration-workflow.yml
oalyman: I remove the docker build for sonarcube

jenkins/master/Dockerfile
oalyman: Yes, works.

PR #584, opened 2020-05-22
Improve project creation scripts

### LINKED ISSUE #333: create-projects: Exchange environment variable with proper command line arguments
Both scripts in create-projects use some undocumented environment variables.

A huge improvement to those scripts would be to exchange the usage of environment variables, which can lead to unforeseen side effects on other scripts or be effected by already exported values, with proper command line arguments with usage documentation.

michaelsauter:
- Cleanup
- Exposing all variables as arguments
- Replaces the ability to run Tailor with `status` by making the script interactive by default, just `--non-interactive` in the `Jenkinsfile` (FYI @henrjk)

Closes #333.

As bonus points, this honours the input from the prov app over the configuration file, allowing to actually check if a differently configured prov app (e.g. for testing purposes) has the desired effect.

I've added some tests against `oc`/`tailor` mocks (based on https://github.com/michaelsauter/bock) to verify at least the correctness of the script logic, as we don't have full end-to-end tests in Github Actions right now ... better than nothing I would say, and hopefully we can write better tests in the future.


create-projects/create-projects.sh
```diff
@@ -1,88 +1,117 @@
...
+  echo "Allow all authenticated users to view the project"
+  oc policy add-role-to-group view system:authenticated -n "${PROJECT_ID}-dev"
+  oc policy add-role-to-group view system:authenticated -n "${PROJECT_ID}-test"
+  oc policy add-role-to-group view system:authenticated -n "${PROJECT_ID}-cd"
+
+  echo "Allow all authenticated users to edit the project"
+  oc policy add-role-to-group edit system:authenticated -n "${PROJECT_ID}-dev"
+  oc policy add-role-to-group edit system:authenticated -n "${PROJECT_ID}-test"
+  oc policy add-role-to-group edit system:authenticated -n "${PROJECT_ID}-cd"
+fi
```
michaelsauter, line 117: The diff looks huge but the changes are very few - I think this is mostly whitespace :(

create-projects/create-projects.sh
```diff
@@ -1,88 +1,119 @@
...
+# for env autoclonding in the Jenkins shared library.
+JENKINS_ROLE="admin"
+
+function usage {
+  printf "usage: %s [options]\n" "$0"
+  printf "\t-h|--help\tPrints the usage\n"
+  printf "\t-v|--verbose\tVerbose output\n"
+  printf "\t-p|--project\tProject ID\n"
+  printf "\t--admins\tAdmins of the projects\n"
+  printf "\t--groups\tGroups with permissions (e.g. 'USERGROUP=foo,ADMINGROUP=bar,READONLYGROUP=baz')\n"
```
michaelsauter, line 17: @clemensutschig Could you confirm whether this is the correct format? I read through the code and this is what I understood from it. Therefore, also the tests assert a value with this format now.

create-projects/create-cd-jenkins.sh
```diff
@@ -1,83 +1,110 @@
...
 
-# This script sets up a ${PROJECT_ID}-cd/jenkins-master and the associated webhook proxy for the project you are creating.
+# This script sets up a ${PROJECT_ID}-cd/jenkins-master and the associated
+# webhook proxy for the project you are creating.
 
 TAILOR="tailor"
-VERBOSE=false
-STATUS=false
+ODS_NAMESPACE="ods"
+ODS_IMAGE_TAG="lastest"
```
henrjk, line 9: There is a typo here.

create-projects/create-cd-jenkins.sh
```diff
@@ -1,83 +1,110 @@
 #!/usr/bin/env bash
 set -eu
 
-# This script sets up a ${PROJECT_ID}-cd/jenkins-master and the associated webhook proxy for the project you are creating.
+# This script sets up a ${PROJECT_ID}-cd/jenkins-master and the associated
+# webhook proxy for the project you are creating.
 
 TAILOR="tailor"
-VERBOSE=false
-STATUS=false
```
henrjk, line 8: I will likely miss the ability to use this with tailor status but it was meant for development mostly so one can also locally change the script.

create-projects/create-cd-jenkins.sh
michaelsauter: Great catch - fixed in e12935f.

create-projects/create-cd-jenkins.sh
michaelsauter: I felt making the script interactive by default would be even more beneficial, as `apply` is the same as `diff` and answering `yes` to the prompt. If you still think you miss `diff`, I can change it back, too. Just thought removing it simplified the script a little, and the script being interactive made up for the loss of the functionality.

create-projects/create-cd-jenkins.sh
```diff
@@ -1,83 +1,110 @@
...
+ODS_NAMESPACE="ods"
+ODS_IMAGE_TAG="lastest"
+PROJECT_ID=""
```
henrjk, line 10: I would not mind having the ability to pass parameters via environment variables so one would not have to explicitly pass them into tailor. 
On the other hand this makes it less explicit on where parameters are used.

michaelsauter: Hmm ... I decided to reset everything as those scripts are executed within the context of Jenkins, which has some env vars exposed (via the `DeploymentConfig`, but also from inside the image). It might be surprising to have them alter what the script does without seeing them passed/set in the `Jenkinsfile`. That's why I opted to have all params explicitly, and prevent any effect of outside env vars.

create-projects/create-projects.sh
```diff
@@ -1,88 +1,119 @@
 #!/usr/bin/env bash
-set -ex
+set -e
+
+PROJECT_ID=""
+PROJECT_ADMINS=""
```
henrjk, line 5: Allowing to pass these as environment variables could make sense as well?

create-projects/tests/run.sh
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env bash
+set -ue
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+PATH=${SCRIPT_DIR}:$PATH
+
+curl --silent -L "https://raw.githubusercontent.com/michaelsauter/bock/master/bock.sh" -o "${SCRIPT_DIR}/oc" && chmod +x "${SCRIPT_DIR}/oc"
```
henrjk, line 7: Could it make sense to not use master but a fixed 'release' of bock.sh?

create-projects/create-cd-jenkins.sh
henrjk: It is ok just wanted to be sure you wouldn't miss it too after all.

create-projects/create-cd-jenkins.sh
henrjk: I can live with that as well. Perhaps we could add a comment in the code so that we don't forget about this later. Or perhaps better add a test that would fail if someone makes changes like that.

create-projects/create-projects.sh
henrjk: This is the same conversation as above. No need to repeat here.

create-projects/create-cd-jenkins.sh
michaelsauter: Cool. Added comments in 1ff8267.

create-projects/tests/run.sh
michaelsauter: Yes - done in 3eda1d7.

PR #587, opened 2020-05-25
changes params in prov app pvc to optional to enable default storage

stitakis:
this fixes broken pvc deployment with tailor. Problem is that when paramerters are set with option `required: true` then Tailor doesn't accept empty values. For ods-devenv we need default values which are normaly set as `empty`.

### DIFF
--- a/nexus/ocp-config/pvc.yml
+++ b/nexus/ocp-config/pvc.yml
@@ -5,10 +5,13 @@ parameters:
   value: nexus
 - name: STORAGE_PROVISIONER
   description: storage class provisioner. For AWS this could be kubernetes.io/aws-ebs. Leave empty for local (e.g. vagrant) deployments
+  value: ""
 - name: STORAGE_CLASS_DATA
   description: storage class for data. For AWS this could be gp2. Leave empty for local (e.g. vagrant) deployments
+  value: ""
 - name: STORAGE_CLASS_BACKUP
   description: Storage class for data backup. For AWS this could be gp2-encrypted. Leave empty for local (e.g. vagrant) deployments
+  value: ""
 - name: NEXUS_DATA_CAPACITY
   value: 60Gi
 - name: NEXUS_BACKUP_CAPACITY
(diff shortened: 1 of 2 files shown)


michaelsauter:
I would prefer to replace `required: false` with `value: ""`, which has the same effect. I am wondering if we need this change elsewhere too? Did you have a look at other templates?

Apart from that, are you sure they need to be empty? Do you know what OpenShift sets those values to if you create a PVC manually in the cluster?

stitakis:
@michaelsauter @georgfedermann I did a search for files named `pvc.xml` in project `ods-core`. There are only 2 files. One is the storage definition of ProvApp. The 2nd corresponds to the Nexus storage definition. The Nexus one doesn't specify the attribute `required` at all. I'm fine with changing it to `value: ""` but I think we should have same definition for same parameters in both `pvc.xml`.

I think, yes for ods dev env we don't need to set any value, defaults assigned by oc are good enough.

This is the output of `oc get pvc prov-history -o json`. It do not contain any storage specific annotation.

```
{
    "apiVersion": "v1",
    "kind": "PersistentVolumeClaim",
    "metadata": {
        "annotations": {
            "pv.kubernetes.io/bind-completed": "yes",
            "pv.kubernetes.io/bound-by-controller": "yes"
        },
        "creationTimestamp": "2020-05-22T16:32:11Z",
        "finalizers": [
            "kubernetes.io/pvc-protection"
        ],
        "name": "prov-history",
        "namespace": "ods",
        "resourceVersion": "40435",
        "selfLink": "/api/v1/namespaces/ods/persistentvolumeclaims/prov-history",
        "uid": "c99dea00-9c49-11ea-b6a8-0a444467b670"
    },
    "spec": {
        "accessModes": [
            "ReadWriteOnce"
        ],
        "resources": {
            "requests": {
                "storage": "100Gi"
            }
        },
        "volumeName": "pv0056"
    },
    "status": {
        "accessModes": [
            "ReadWriteOnce",
            "ReadWriteMany",
            "ReadOnlyMany"
        ],
        "capacity": {
            "storage": "100Gi"
        },
        "phase": "Bound"
    }
}
```

I'd go for `required: ""` in ProvApp and Nexus `pvc.xml`.

michaelsauter:
OK, please update both then to use `value: ""`. However, I would assume that also the Jenkins template must have that as it uses PVCs. Same for SonarQube (Postgres)?

PR #592, opened 2020-05-26
change ca cert importing for jenkins and sonarcube to runtime

oalyman:


jenkins/slave-base/Dockerfile
```diff
@@ -86,6 +69,8 @@ RUN if [ -z $SNYK_DISTRIBUTION_URL ] ; then echo 'Skipping snyk installation!' ;
 
 # Set java proxy var.
 COPY set_java_proxy.sh /tmp/set_java_proxy.sh
+RUN cp /usr/local/bin/run-jnlp-client /usr/local/bin/ocp-run-jnlp-client
+COPY ods-run-jnlp-client /usr/local/bin/run-jnlp-client
```
michaelsauter, line 73: Should we ensure here that `/usr/local/bin/run-jnlp-client` is executable? I think elsewhere we do that consistently. Maybe you can use this to also fix the next line - instead of `chmod 777` a simple `chmod +x /tmp/set_java_proxy.sh` should do.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -1,6 +1,18 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+      -
+        name: Checkout repository
+        uses: actions/checkout@v2.0.0
+      -
+        name: Build image
+        working-directory: jenkins/master
+        run: |
+          docker build -t jenkins .
```
michaelsauter, line 15: Do you mind changing this job to `jenkins-master`, and adding another job `jenkins-slave` next to it that tests the slave? It would be really nice to see that at least building the image works.

PR #594, opened 2020-05-27
update nexus to 3.23.0

cschweikert:
security patch for CVE-2020-11753. See also https://support.sonatype.com/hc/en-us/articles/360046233714


michaelsauter:
@cschweikert Could we go to 3.23.0 as well or any reason not to do that? Given that we are on `master` I think we can stay with the latest version if possible.

cschweikert:
Good point. No breaking changes as far as I can see from the release notes... and ``npm audit`` will finally work :)

PR #549, opened 2020-05-12
Basic setup of a CentOS 7 machine as an ODS dev environment

georgfedermann:
Closes #398

Status:
- Atlassian Jira and BitBucket using Atlassian timebomb license
- MySQL 5.7
- Docker infrastructure
- SSH / RDP infrastructure incl firewall configuration
- OpenShift 3.11 cluster
- Tailor 0.13.1
- ODS repositories in local BitBucket instance
- Tested on VMWare Fusion and AWS EC2
Next steps will be:
- Integration ODS in OpenShift cluster incl. SonarQube, Nexus
- Setup Crowd 
- Integration provisioning-app
Merging this PR would allow to fix a reference here: https://github.com/opendevstack/ods-core/blob/153f6e1a1622322b081d461441a714b517dc9613/ods-devenv/scripts/deploy.sh#L432

PR #600, opened 2020-05-29
Running E2E tests from local machine using AWS infrastructure

hugowschneider:
This is a Work In Progress for fixing the E2E test code and create a script for running the test in an AWS environment automatically from a local development environment.


clemensutschig:
@hugowschneider  - how do we ensure there is as much as possible reuse from @georgfedermann 's work .. (largely PR: #549)

hugowschneider:
@clemensutschig, the only focus here is to run the old tests again using VMs in AWS with the help with @michaelsauter.  That means we are trying run what is defined in the GitHub Actions workflow.

michaelsauter:
@clemensutschig I was going through this with @hugowschneider earlier today.

I see it like this:

1. Get the tests back to a working state (executed manually on demand)
2. Make use of the scripts behind the `Makefile` targets to create the required OpenShift env to run the tests in. Those scripts are also used in #549. So that is how the reuse happens.
3. Extract anything that is still custom in the test setup and #549 into common scripts, which then also get a "Makefile interface" so that they can be run by admins (if applicable)

Apart from that, there are a couple of differences between this and #549 

- The dev VM should use the real software pieces, the tests mock most things (right now)
- The existing tests do not cover things end-to-end but serve as a smaller integration test

clemensutschig:
I am totally happy .. :) just asking

hugowschneider:
@clemensutschig, not yet. Some tests still need fixing and refactoring and the script to run the test in AWS is not ready also.

michaelsauter:
@hugowschneider I think this should be merged now. It is in a state where one can reliable start an EC2 instance and execute tests in it.

That said, the tests fail right now:

* There's a diff on the `jenkins_edit` role, which is why the check is commented out. It appears that the `RoleBinding` creates a `ClusterRole`, whereas doing the same against a cluster that I have access to, it is a normal `Role`. Maybe it depends how executes it? The tests use an admin. I tried to force the role in the template, but to no avail.
* Running the `Jenkinsfile` doesn't work as the application is down

Even with failing tests I would want to merge this, as now everyone should be able to look at the problems and fix it, whereas without this PR you have to figure out lots by yourself.

hugowschneider:
@michaelsauter Everything looks good and I agreed we should merge, and PR each test fix later on.

PR #601, opened 2020-05-29
removes clone-script involvement in credential handling

henrjk:
This is to be delivered with https://github.com/opendevstack/ods-jenkins-shared-library/pull/363.

Credential handling is now expected to be supplied external to the script, namely interactively/OS integration.

This can be done by having a netrc file in place or using a git credential handler.

This also removes code for backwards compatibility with pre 2.x ods-jenkins-shared-library versions.

Also adds some checking of assumptions made in ods-jenkins-shared-library about the names of the scripts so that inadvertent changes in naming should be caught. Also extends some shellcheck testing. 

See https://github.com/opendevstack/ods-jenkins-shared-library/issues/112#issuecomment-629282926 for additional context.

### DIFF
--- a/ocp-scripts/verify-jsl-expectations.sh
+++ b/ocp-scripts/verify-jsl-expectations.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# This script verifies that the clone-project.sh and the export-/import
+# scripts used by it are not inadvertantly renamed.
+# ods-jenkins-shared-library depends on this.
+
+# The script is expected to be exucuted from the project root.
+
+set -e
+
+script_names=(clone-project.sh export-project.sh export-project.sh)
+
+for name in "${script_names[@]}"; do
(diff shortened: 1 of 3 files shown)


ocp-scripts/verify-jsl-expectations.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# This script verifies that the clone-project.sh and the export-/import
+# scripts used by it are not inadvertantly renamed.
+# ods-jenkins-shared-library depends on this.
+
+# The script is expected to be exucuted from the project root.
+
+set -e
+
+script_names=(clone-project.sh export-project.sh export-project.sh)
```
gerardcl, line 11: Hi @henrjk , sorry for my delayed review!
question: isn't import-project.sh missing but export one duplicated in the list?

PR #603, opened 2020-06-02
Webhook proxy resource constraints

michaelsauter:
Noticed there were no constraints either. Given my research that should be the last instance that did not specify any constraints.

### DIFF
--- a/jenkins/ocp-config/deploy/jenkins-webhook-proxy.yml
+++ b/jenkins/ocp-config/deploy/jenkins-webhook-proxy.yml
@@ -18,6 +18,14 @@ parameters:
   required: true
 - name: PIPELINE_TRIGGER_SECRET_B64
   required: true
+- name: WEBHOOK_PROXY_CPU_REQUEST
+  value: 50m
+- name: WEBHOOK_PROXY_CPU_LIMIT
+  value: 200m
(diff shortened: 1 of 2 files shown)


jenkins/ocp-config/deploy/jenkins-webhook-proxy.yml
```diff
@@ -18,6 +18,14 @@ parameters:
   required: true
 - name: PIPELINE_TRIGGER_SECRET_B64
   required: true
+- name: WEBHOOK_PROXY_CPU_REQUEST
+  value: 50m
+- name: WEBHOOK_PROXY_CPU_LIMIT
+  value: 200m
```
henrjk, line 24: Whenever I look at the webhook proxy metrics it appears to have even lower CPU needs. Perhaps 100m max?

michaelsauter: Hmm yea tried to be not so strict here, to help during boot time. Usually we have a 1:10 ratio for CPU request/limit, but that seemed excessive in this case. So I ended up opting for 200m (1:4 ratio). In the end, I think overcommitting the cluster CPU wise is not as bad as doing that for memory ...

PR #611, opened 2020-06-04
add original certificate importing for jenkins master, slave-base and sonarcube

oalyman:


sonarqube/docker/Dockerfile
```diff
@@ -34,6 +35,15 @@ RUN set -x \
     && mkdir -p /opt/configuration/sonarqube \
     && chown -R sonarqube:root /opt/configuration/sonarqube
 
+# fetch certificates and store them in tmp directory
+RUN if [ -z $idpDns ]; then echo 'Skipping custom certificate!'; else \
```
clemensutschig, line 39: why not switching this to APP_DNS?

clemensutschig: and also we have introduce APP_DNS_PORT ... because clusters (could) run on a different port

clemensutschig:
@oalyman  - here is my proposal- I know this is slightly larger - but will serve great good

1. Introduce a dockerfile in ODS core for each component (SQ / PROV APP / DOC GEN) - that does not much more than `FROM opendevstackorg/...` and then sets up the cert / http proxy mess

1. harmonize to APP_DNS / APP_DNS_PORT - vs. idp ..

1. Change the ods core BCs to point to those dockerfiles ... 

cool? @michaelsauter fyi

clemensutschig:
@oalyman cool?

PR #641, opened 2020-06-22
Rename jenkins-slave-base to jenkins-agent-base

### LINKED ISSUE #633: Rename jenkins-slave-base to jenkins-agent-base
I propose to do this for v3, as we change the central namespace from `cd` to `ods`, so people need to update the image stream tag anyway.

OpenShift offers "agent images" for a while now for v3 (https://docs.openshift.com/container-platform/3.10/using_images/other_images/jenkins.html): 

> the jenkins-slave-maven-* and jenkins-slave-nodejs-* images are being marked as deprecated during the v3.10 release cycle. The images will still exist in the interim so users can migrate their applications to the newer jenkins-agent-maven-* and jenkins-agent-nodejs-* images.

Jenkins also renamed their images recently: https://www.jenkins.io/blog/2020/05/06/docker-agent-image-renaming/.

As a follow-up, we also need to rename the Jenkins images in `ods-quickstarters`.

FYI @oalyman @bljubisic @clemensutschig @borja44

michaelsauter:
Closes #633.

We should probably wait with merging until we have a corresponding PR for the quickstarters as well, as they would break with this PR ... @oalyman @bljubisic Could one of you provide this PR?


oalyman:
@michaelsauter I can provide the PR

michaelsauter:
FYI @georgfedermann This also changes code in the devenv scripts. When we merge this, the code will be broken until we merge the follow-up PR in the ods-quickstarters repo, should be coming right after this one is merged.

PR #642, opened 2020-06-22
Fix Nexus setup

michaelsauter:
A few adjustments based on what I ran into in a recent installation.

@georgfedermann I have two questions:

* Did you find anything regarding the `npm-all` repo? I could not find it so I think we can remove it
* Did `oc rsh rm ...` work in the EDP box? For me, it worked but returned exit code 129, so I had to "resume" the script ...


georgfedermann:
> * Did you find anything regarding the `npm-all` repo? I could not find it so I think we can remove it
We did not find any usage and did not find anyone using npm-all, but it might still be used by some project out there ...
> * Did `oc rsh rm ...` work in the EDP box? For me, it worked but returned exit code 129, so I had to "resume" the script ...
So far we did not do scripted access to OpenShift containers with oc rsh in the EDP box.

michaelsauter:
> We did not find any usage and did not find anyone using npm-all, but it might still be used by some project out there ...

That is fine, as here we are only changing the *setup* of Nexus, not an existing installation.

> So far we did not do scripted access to OpenShift containers with oc rsh in the EDP box.

If you run `make configure-nexus` (and you should?), then you do run the snippet `oc rsh rm ...` inside the script.

clemensutschig:
@michaelsauter  - https://github.com/opendevstack/ods-core/blob/master/ods-devenv/scripts/deploy.sh#L1279-L1287
 we use the make approach ...

PR #643, opened 2020-06-23
add smoke tests

georgfedermann:


tests/README.md
```diff
@@ -1,37 +0,0 @@
...
-Don't forget to terminate the instance once you are done testing.
-
-Often, you'd want to run the tests against the local state (e.g. to check if
-your modifications break any tests). To do this, pass `--rsync` to
-`./run-on-aws.sh`.
-
-## General Usage
-
-You can also run the tests outside AWS. Just execute `install.sh`,
-`prepare-test.sh` and `test.sh` on any Ubuntu 16.04 host.
```
michaelsauter, line 37: @georgfedermann I would love to have the contents of this file somewhere. Right now I am a bit lost how to run "EDP in a box" on AWS. I can see some instructions at https://github.com/opendevstack/ods-core/tree/master/ods-devenv, and there is a script at https://github.com/opendevstack/ods-core/blob/master/ods-devenv/scripts/run-on-aws.sh. They don't seem to be in sync? Which one should I use? And is the pre-baked AMI available everywhere?

Further, I was wondering why https://github.com/opendevstack/ods-core/blob/master/tests/install.sh was left in the tree? Is it still used now?

PR #655, opened 2020-06-26
update admin update guide

oalyman:


docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+git reset --hard ods/master
+git push origin master
+# Update 2.x
+git checkout 2.x
+git reset --hard ods/2.x
+git push origin 2.x
+----
+
+If your OpenDevStack installation is based on a custom branch (such as `2.acme`), then you
+need to create a pull request on BitBucket from `2.x` into that custom branch now.
```
michaelsauter, line 37: I think pretty much all of the above can be replaced by `make sync-repos` (which asks which revision to use).

For custom branch, people need to run sth like `ods-setup/repos.sh --source-git-ref master --target-git-ref 3.x`

docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+== Tailor
+
+Next, update Tailor.
+3.x requires Tailor https://github.com/opendevstack/tailor/releases/tag/v1.1.1[1.1.1].
+
+== Configuration
+
+Then, update/add/remove the configuration parameters (located in `ods-configuration`).
+To do this, use the `./update` script located in `ods-core/configuration-sample`.
```
michaelsauter, line 69: Should be `make prepare-config`

docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+== OCP resources
```
michaelsauter, line 71: Can we use `OpenShift` instead of `OCP` everywhere? I think `OCP` is a specific product of RedHat.

docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+Next, run `tailor update` in `ods-core` and `ods-quickstarters` to bring all OCP resources (such as DC or BC) into sync. Review the diff produced by Tailor carefully, especially around changes to PVCs.
```
michaelsauter, line 73: Please update to `tailor apply` instead of `tailor update`. However, people should probably run the `Makefile` targets from the installation section.

docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+
+== Images
+
+After all OCP resources have been updated, you need to start a build for all build configs
+in the `cd` namespace to create new images.
```
michaelsauter, line 78: Should be `ods`, but I think this section is not needed at all if you run the `Makefile` targets.

docs/modules/administration/pages/update-2-to-3.adoc
```diff
@@ -0,0 +1,137 @@
...
+tailor export -n prov-dev > backup_PROV_DEV.yml
+tailor export -n prov-test > backup_PROV_TEST.yml
+----
+
+Note that the executing user needs to have permissions to access all resources
+in the `cd` namespaces for this to work properly.
+
+== Tailor
+
+Next, update Tailor.
```
michaelsauter, line 63: Would be nice to link to the install instructions here.

clemensutschig:
@oalyman  - do you fix this? just curious ... :D

PR #657, opened 2020-06-26
sets default profile as active profile in prov app

stitakis:


stitakis:
fixes https://github.com/opendevstack/ods-provisioning-app/issues/477

Issue #411, opened 2020-02-10
Jenkins plugins are not updated in existing pods


stefanlack:
**Describe the bug**
When a new jenkins master image is build with updated plugin versions and the image is deployed to the project, the new plugin versions are not used. Instead, the old plugin versions are loaded

**To Reproduce**
1. Update plugins.txt to use new versions
2. Build jenkins master in global cd project
3. Wait until trigger redeploys jenkins master in a project specific cd - namespace
4. Login to jenkins and verify, that plugin has expected version

**Expected behavior**
Version correct.

**Affected version (please complete the following information):**
 - OpenShift:  3.11
 - OpenDevStack 1.2, 2

**Additional context**
To solve the issue, one have to set the environment variable `OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS=true`

See https://github.com/openshift/jenkins#environment-variables for details

Also `OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG=true` may be needed.

One problem with this approach is, that jenkins looses history after restart.

Does you also examined this problem and how to you work around the problem?

michaelsauter:
I think we face the same issues. Not sure how to tackle this right now.

gerardcl:
Question: do we really need a PVC? shall we just make Jenkins stateless? If restarted it will anyway recover the jobs required from openshift pipelines.
What is the real need to having `/var/lib/jenkins` persisted?

michaelsauter:
Hmm, good question. I think build numbers would be affected at least?

gerardcl:
that is true, the counter is restarted
@clemensutschig do you see an issue on this? or any other?

michaelsauter:
Note that this ONLY affects plugins in `plugins.txt`. Plugins shipped in the RedHat-distributed image do get updated.

michaelsauter:
Setting `OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS` allows to just delete / overwrite plugins without loosing the whole history: https://github.com/openshift/jenkins/blob/ae52cc141e027903576e71f3c038260a3599b82f/2/contrib/s2i/run#L499-L505

clemensutschig:
@michaelsauter I remember you fixed this at bi x.. adding to ods 3?

michaelsauter:
@clemensutschig Still in the process. The whole situation is a bit complex:

* RedHat and CentOS based images differ in how they handle plugins
[...]

PR #672, opened 2020-07-02
Fix Jenkins Master

### LINKED ISSUE #670: Latest jenkins-master:v3.11 breaks
Latest Jenkins Master image (`v3.11.232`) from RedHat ships with Jenkins 2.222 and a bunch of new plugins. Updating to that breaks the pipelines, as new pods do not launch. The reason is that the cloud configuration gets lost somehow ... adding it back manually fixes the issue.

Workaround for now is to pin the tag to `v3.11.219`.

michaelsauter:
Fixes #670.

Solution implemented and tested with @oalyman. We tried creating a new Jenkins from scratch and updating a existing Jenkins (2.x based and master based).

Also:
* Enforces plugin versions of base image (partially fixing #411, the plugins defined in "our" `plugins.txt` still don't get updated)
* Fixes a bug in the proxy configuration (missing import)

### DIFF
--- a/docs/modules/jenkins/examples/plugins.txt
+++ b/docs/modules/jenkins/examples/plugins.txt
@@ -1,11 +1,4 @@
-token-macro:2.3
 greenballs:1.15
-credentials-binding:1.18
-email-ext:2.63
+email-ext:2.69
 sonar:2.6.1
-ansicolor:0.5.2
-workflow-cps-global-lib:2.13
-junit:1.28
-blueocean:1.18.0
-workflow-durable-task-step:2.28
-openshift-sync:1.0.44
+ansicolor:0.7.0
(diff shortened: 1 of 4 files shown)


michaelsauter:
I need to revisit this. We shipped with `blueocean:1.18.0`, but the base image ships with `blueocean:1.10.2`. Unfortunately the logic in place here means that even if we define a higher version in our `plugins.txt`, the installed version is the one from the base image :( We need the "full" solution which sets all plugins defined by either our image or the base image exactly to the versions defined in the images, regardless of what the user did.

clemensutschig:
I think we should use the install_ piece .. that seems to take care, no?

michaelsauter:
Superseded by #701.

PR #674, opened 2020-07-03
aim to limit restarts of nexus

henrjk:
Clemens asked me to look into reducing the number of restarts.
If the statements below are correct this should be mergeable. 

When nexus runs externally we assume there is no initial wait needed.

With docker:
- Changing the ScriptSettings can be done without requiring the server to be up.
- There should be no need to set the script setting to false again.
- This results in only one time waiting for the local case.

### DIFF
--- a/nexus/configure.sh
+++ b/nexus/configure.sh
@@ -188,7 +188,10 @@ function changeScriptSetting {
     fi
 }
 
-# changeScriptSetting does not require server to be up.
+# In the local context changeScriptSetting works directly after docker run
+if [ -z "${LOCAL_CONTAINER_ID}" ]; then
+    waitForReady
+fi
 changeScriptSetting "true"
 
 waitForReady
--- a/nexus/configure.sh
+++ b/nexus/configure.sh
@@ -188,8 +188,7 @@ function changeScriptSetting {
     fi
 }
 
-waitForReady
-
+# changeScriptSetting does not require server to be up.
 changeScriptSetting "true"
 
 waitForReady
@@ -264,6 +263,10 @@ sed "s|@developer_password@|${DEVELOPER_PASSWORD}|g" json/developer-user.json > 
 runJsonScript "createUser" "-d @json/developer-user-with-password.json"
 rm json/developer-user-with-password.json
 
-changeScriptSetting "false"
+if [ -z "${LOCAL_CONTAINER_ID}" ]; then
+    changeScriptSetting "false"
+    waitForReady
+else
+    true # on local docker nexus don't care about security
+fi
 
-waitForReady


nexus/configure.sh
```diff
@@ -188,8 +188,7 @@ function changeScriptSetting {
     fi
 }
 
-waitForReady
-
+# changeScriptSetting does not require server to be up.
```
michaelsauter, line 191: Good point. However, it requires the container to be started, which might not always be the case - e.g. if image is pulled it might take longer then the script? I think we must at least ensure that we can `rsh`/`exec` into the container.

clemensutschig: in https://github.com/opendevstack/ods-core/blob/5c6dce3d070746bf36497b299122ea09b75c4584/nexus/configure.sh#L194 - there is waitForready .. so should be all good, or?

michaelsauter: https://github.com/opendevstack/ods-core/pull/674/files#diff-475161238d774e63be4192b439d0bafdR175 That is executed before line 194, and I'm concerned this will just fail the script depending on timing.

clemensutschig: @michaelsauter  - I guess you are worried over 
```
changeScriptSetting "true"
```
in line 191 ... right?

henrjk: Based on your question I tried to find some docs on when `docker run` in detached mode would return. While I did not find this documented or commented on, I also tried twice to run the script when the image is not in the local cache. This looks like this:

```bash
nexus $ ./test.sh -v
...
+ echo 'Run container using image sonatype/nexus3:3.22.0'
Run container using image sonatype/nexus3:3.22.0
++ docker run -d -p 8081:8081 -e HTTP_PROXY= -e HTTPS_PROXY= -e NO_PROXY= sonatype/nexus3:3.22.0
Unable to find image 'sonatype/nexus3:3.22.0' locally
3.22.0: Pulling from sonatype/nexus3
0bb54aa5e977: Already exists
941e1e2b31a8: Already exists
fb7cd7cf24c9: Pulling fs layer
4e63afcbf773: Pulling fs layer
fb7cd7cf24c9: Download complete
fb7cd7cf24c9: Pull complete
4e63afcbf773: Download complete
4e63afcbf773: Pull complete
Digest: sha256:81d182285d279081e80e74dbd13cb544fdf4255efadd61321436a577f56b87ad
Status: Downloaded newer image for sonatype/nexus3:3.22.0
+ containerId=80a023021605ae245e874b03c4d10985d48c06b14b28436612ed503688a4c3bf
+ trap cleanup EXIT
+ NEXUS_URL=http://localhost:8081
+ NEXUS_ADMIN_USERNAME=admin
+ NEXUS_ADMIN_PASSWORD=s3cr3t
+ NEXUS_USERNAME=developer
+ NEXUS_PASSWORD=geHeim
+ echo 'Run ./configure.sh'
Run ./configure.sh
```  
From this it seems that it will not return before the images are pulled. Also to have a containerId it may need to have it started already (not sure either about this).

So the current code did not fail for me.

I am hesitating to put in a wait loop for being able to `docker exec` in the container because of these observation. @michaelsauter  what do you think?

clemensutschig: agreed ... so good to go from my side!

michaelsauter: @clemensutschig I wasn't thinking about `docker run`, that will always work as it doesn't detach before the container is started. I was wondering about the OpenShift case, which is async - the pulling happens after `oc apply` ran, so potentially it is not up. I would keep the first check in - it doesn't take up time anyway if the service is already running.

PR #677, opened 2020-07-03
ods-verify: parameterize bitbucket project and ocp one

clemensutschig:
take the config for bitbucket project & ocp one - rather than hardcode


clemensutschig:
@georgfedermann  - feel free to try on ods in a box
@michaelsauter  - I took the freedom now to rename the tests - to spell out what they are doing :D

PR #685, opened 2020-07-06
Feature/add jenkins json golden check

clemensutschig:


tests/scripts/utils/print-jenkins-log.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-LOG_URL=$(oc get build $1 -o json | jq '.metadata.annotations."openshift.io/jenkins-log-url"' | sed -e 's/"//g' )
+LOG_URL=$(oc get build $1 -o json | jq '.metadata.annotations."openshift.io/jenkins-log-url"' | sed 's/"//g' )
```
michaelsauter, line 3: You can use `jq -r` to avoid `| sed 's/"//g'` (also above)

Issue #680, opened 2020-07-05
Specify which verison of Git repos.sh supports


sino92:
**Is your feature request related to a problem? Please describe.**
While running the script on our base Jenkins agent, I've noticed it wasn't properly executing a boolean statement [here](https://github.com/opendevstack/ods-core/blob/master/ods-setup/repos.sh#L191). Because the statement ended up with `false`, it was trying to reassign the already existing `ods` remote. It wasn't clear from the logs as `git remote get-url ods` standard output was suppressed.    

On another note, I believe OC centos base images ship with git version `1.8.3.1`. I imagine there are some deprecated methods since that release but it would be nice to use later versions of Git in our images.

**Describe the solution you'd like**
Clarify which version of GIT the script supports. In this [Git Release Note](https://github.com/git/git/blob/228f53135a4a41a37b6be8e4d6e2b6153db4a8ed/Documentation/RelNotes/2.7.0.txt#L13), you can see that get-url feature was added for version 2.7. It'd be extremely helpful to see that noted in the script. 

**Describe alternatives you've considered**
Don't suppress the output of `git remote get-url ods`?

michaelsauter:
Requirements are listed here: https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_server_prerequisites.

Where / How would you like to see this with `repos.sh`?

sino92:
Ah its cool we already have the server requirements. It sounds like we need git version `>= 2.13.2` only for OC and Atlassian Suite. One can misunderstand then `repos.sh` doesn't adhere to such system requirements. In my use case, I've chosen to clone ODS codebase in a separate machine (that doesn't run OC or Atlassian) to regularly update a remote Bitbucket ODS fork. 

Maybe we can add the git version requirement to the same document and to the script itself?

michaelsauter:
@sino92 You're right, the docs are not clear that this is a "ODS prerequisite". Do you mind updating it?

sino92:
Sure, will do.

PR #690, opened 2020-07-06
Specify requirements for repos.sh

sino92:
https://github.com/opendevstack/ods-core/issues/680

### DIFF
--- a/ods-setup/repos.sh
+++ b/ods-setup/repos.sh
@@ -27,6 +27,7 @@ BITBUCKET_ODS_PROJECT="opendevstack"
 SYNC="n"
 
 function usage {
+  printf "NOTE: Before you run this script, make sure to have at least Git 2.7 in your system.\n\n"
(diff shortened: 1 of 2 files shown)


ods-setup/repos.sh
```diff
@@ -27,6 +27,7 @@ BITBUCKET_ODS_PROJECT="opendevstack"
 SYNC="n"
 
 function usage {
+  printf "NOTE: Before you run this script, make sure to have at least Git 2.7 in your system.\n\n"
```
michaelsauter, line 30: I think it would be nicer to have the same Git version everywhere to avoid confusion. So even though this may run with 2.7, let's require 2.13 everywhere so people update their Git installations. 2.13 was released May 2017, so is already over 3 years old.

sino92: True, better to be consistent and going for the higher version will be safer

Issue #681, opened 2020-07-05
Specify which service versions ODS officially supports


sino92:
An example of what it would look like is [here](https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L75). We can add supported versions to this page and the front page of ODS documentation to increase visibility.

sino92:
@michaelsauter has addressed this issue in this commit https://github.com/opendevstack/ods-core/commit/cc3734065fa25455e7affd060f83cdf2ace6dd1a

PR #693, opened 2020-07-06
Specify service requirements for certain tools

sino92:
#681

### DIFF
--- a/configuration-sample/ods-core.env.sample
+++ b/configuration-sample/ods-core.env.sample
@@ -130,7 +130,7 @@ CONFLUENCE_URL=http://192.168.56.31:8090
 # Jenkins #
 ###########
 
-# Base image for Jenkins master (v2.176.3).
+# Base image for Jenkins master.
 # For RHEL7, you can use e.g. registry.access.redhat.com/openshift3/jenkins-2-rhel7:v3.11
 # See available tags for RHEL7-based base images: https://catalog.redhat.com/software/containers/openshift3/jenkins-2-rhel7/581d2f4500e5d05639b6517b
 # See available tags for CENTOS7-based base images: https://hub.docker.com/r/openshift/jenkins-2-centos7/tags
@@ -155,7 +155,7 @@ SHARED_LIBRARY_REPOSITORY=https://github.com/opendevstack/ods-jenkins-shared-lib
 # OpenShift #
 #############
 
-# Openshift v3.11.154
+# Openshift v3.11
 # Internal docker registry host and port - this is used 
 # for pulling the agent images in jenkins.
 DOCKER_REGISTRY=docker-registry.default.svc:5000
--- a/configuration-sample/ods-core.env.sample
+++ b/configuration-sample/ods-core.env.sample
@@ -130,7 +130,7 @@ CONFLUENCE_URL=http://192.168.56.31:8090
 # Jenkins #
 ###########
 
-# Base image for Jenkins master.
+# Base image for Jenkins master (v2.176.3).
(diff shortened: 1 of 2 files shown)


configuration-sample/ods-core.env.sample
```diff
@@ -155,6 +155,7 @@ SHARED_LIBRARY_REPOSITORY=https://github.com/opendevstack/ods-jenkins-shared-lib
 # OpenShift #
 #############
 
+# Openshift v3.11.154
```
michaelsauter, line 158: Hmm is that the latest? I would note down maybe what is tested. But keep in mind that we need to constantly update this. I would avoid for OpenShift to be more specific than 3.11.

docs/modules/administration/pages/installation.adoc
```diff
@@ -13,6 +13,16 @@ In order to run the Atlassian suite and OpenShift, your host must have:
...
 
+== Service Requirements
+ODS 3.x officially supports:
+- Openshift v3.11.154. The supported images for this version of Openshift can be found at:
+    - https://docs.openshift.com/container-platform/3.11/release_notes/ocp_3_11_release_notes.html
+- Jenkins v2.176.3
+    - The base image of the Jenkins we use through this version of ODS is: `openshift3/jenkins-2-rhel7:v3.11.153-2`
+- Sonarqube 
+    - 7.9 (LTS release)
+    - 8.2.0.32929 (latest release)
```
michaelsauter, line 24: I would just add a comment in the config step to tell people to look at the supported / tested versions when they configure ... and the actual versions are in the sample. I just want to avoid having all sorts of numbers in various places that do not get updated in sync ;)

sino92: Fair enough, keeps it simpler to maintain

configuration-sample/ods-core.env.sample
sino92: You're right, I didn't notice the bugfix updates released every month or so. Makes sense that keeping at 3.11 is enough for now.

configuration-sample/ods-core.env.sample
```diff
@@ -130,7 +130,7 @@ CONFLUENCE_URL=http://192.168.56.31:8090
 # Jenkins #
 ###########
 
-# Base image for Jenkins master.
+# Base image for Jenkins master (v2.176.3).
```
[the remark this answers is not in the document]
sino92: I got that version from one of the existing project instances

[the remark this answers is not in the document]
michaelsauter: I see - that is not the latest one then. v3.11.219 is known to work with ODS, and I'm trying to fix the issues with v3.11.232 at the moment.

michaelsauter:
@sino92 Now we have merge conflicts :( To make things easier, I just went ahead and made some adjustments directly, see https://github.com/opendevstack/ods-core/commit/cc3734065fa25455e7affd060f83cdf2ace6dd1a.

The only thing that you also have in this PR is the comment about checking the sample file before starting the config. If you feel that is needed, please update this PR accordingly. Personally, I don't feel it is that necessary as, when you run `prepare-config`, you are creating an `ods-core.env` file with all the comments from the sample file. So when people fill out that file, they see the comments anyway ...

What do you think? Do you want to close or update the PR?

sino92:
@michaelsauter I really liked the changes in this commit `cc37340`. Looks much better with the tag specified. 
I hadn't caught prepare-config before, yes that seems like enough for me tbh. We can close this PR and the issue.

PR #694, opened 2020-07-07
Log non-successful return code when triggering pipeline

michaelsauter:


michaelsauter:
@clemensutschig It should, yes. https://github.com/opendevstack/ods-core/pull/694/files#diff-e133f8a1e377f52183f17a446d708263R462-R465 makes the proxy return the status code that it received from OpenShift. https://github.com/opendevstack/ods-core/pull/694/files#diff-322f82861a2c48f4c2f605f1640c699bR552-R566 adds an example test for this.

PR #701, opened 2020-07-09
Fix Jenkins master issues 

### LINKED ISSUE #411: Jenkins plugins are not updated in existing pods

### LINKED ISSUE #670: Latest jenkins-master:v3.11 breaks

michaelsauter:
Fixes #670 by ensuring that all plugins defined by the RedHat-provided base image are actually deployed in the version defined there. Previously, some plugins would not have been updated in older Jenkins instances (because the plugins on the volume were not symlinked, but overwritten by our `plugins.txt`).

Fixes #411 by copying all plugins either defined in the RedHat-provided base image (see above) OR installed via our `plugins.txt` to the volume on every subsequent Jenkins boot. As a consequence, users of ODS cannot change the plugin versions for those plugins long-term. Every container boot resets the plugin version.

Also, it fixes a bug introduced by #684, as Groovy is tripped up by the space in the command line argument. Luckily, Go templates also work without a space between `if .spec.tls`.

And finally, it removes a bunch of plugins from `plugins.txt` that are anyway provided by the base image. Only exception is `blueocean`, which is actually provided in version 1.10. However, we need a newer version, so we define 1.18 in our `plugins.txt` (and that is the version which actually gets installed).

To test this, I've updated several Jenkins instances in various states (2.176, 2.204, 2.222) as well as updating some plugins manually and restarting. I hope that I covered all edge cases but this whole thing is very brittle and once RedHat ships new base images we need to verify everything again :(

PR replaces #672.

### DIFF
--- a/jenkins/master/configuration/init.groovy.d/proxy.groovy
+++ b/jenkins/master/configuration/init.groovy.d/proxy.groovy
@@ -6,14 +6,14 @@ Jenkins jenkins = Jenkins.getInstance()
 String httpProxy = System.getenv('HTTP_PROXY')
 
 if (!httpProxy) {
-  println("No proxy config found - returning..")
+  println("No proxy config found")
   return
 }
 
 if (jenkins.proxy) {
-  println("Proxy ALREADY set to: ${jenkins.proxy.name}:${jenkins.proxy.port} - leaving ..")
+  println("Proxy ALREADY set to: ${jenkins.proxy.name}:${jenkins.proxy.port}")
   return
-} 
+}
 
 httpProxy = httpProxy.minus(~/^https?:\/\//)
 println (httpProxy)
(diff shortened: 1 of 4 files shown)


clemensutschig:
wow ... this is super cool, well done @michaelsauter

PR #719, opened 2020-07-22
Read ODS_NAMESPACE from config

michaelsauter:
It seems that the "multiple ODS installations in one cluster" use case is more prominent than we originally anticipated. To make usage of it easier, the `Makefile` targets now read the namespace from the config, which avoids having to pass the namespace all the time like `make NAMESPACE=abc install-jenkins`.

Further:
* extract reading a param value from the config into its own script so that it can easily be reused
* reading the Bitbucket Project from config as well

FYI @renedupont @segator pay attention to this when installing next time :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,9 +4,9 @@ SHELL = /bin/bash
 MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
-NAMESPACE=ods
-NEXUS_URL=
-SONARQUBE_URL=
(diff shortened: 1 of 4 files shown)


clemensutschig:
@segator  - you'll need to configure it .. e.g. https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L49

segator:
you are right https://github.com/opendevstack/ods-core/blob/master/sonarqube/ocp-config/sonarqube.yml#L433

michaelsauter:
The routes need to be configured manually. They are not auto-configured based on the namespace. In theory, this could be automated so we can potentially revisit this for v4.

PR #727, opened 2020-07-27
removed/replaced redundant PROV_APP_PIPELINE_TRIGGER_SECRET

### LINKED ISSUE #722: PROV_APP_PIPELINE_TRIGGER_SECRET seems to be redundant
The property [PROV_APP_PIPELINE_TRIGGER_SECRET](PROV_APP_PIPELINE_TRIGGER_SECRET) seems to be redundant, since there is already [PIPELINE_TRIGGER_SECRET](PIPELINE_TRIGGER_SECRET) 

- Remove PROV_APP_PIPELINE_TRIGGER_SECRET  from https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L276
- Replace PROV_APP_PIPELINE_TRIGGER_SECRET by PIPELINE_TRIGGER_SECRET in https://github.com/opendevstack/ods-core/blob/master/ods-provisioning-app/ocp-config/cm.yml#L59 and https://github.com/opendevstack/ods-core/blob/master/ods-provisioning-app/ocp-config/cm.yml#L208

renedupont:
fixes #722

### DIFF
--- a/configuration-sample/ods-core.env.sample
+++ b/configuration-sample/ods-core.env.sample
@@ -272,9 +272,6 @@ PROV_APP_LOG_LEVEL_ATLASSIAN_CROWD=INFO
 PROV_APP_LOG_LEVEL_OPENDEVSTACK=INFO
 PROV_APP_LOGGING_FILE_PATH=/opt/provision/history/logs
 
-# Jenkins pipeline trigger secret (raw)
-PROV_APP_PIPELINE_TRIGGER_SECRET=changeme
-
 # Comma-separated list of events for which Bitbucket sends a request to the webhook proxy service
 # to create/start a Jenkins pipeline or delete a Jenkins pipeline.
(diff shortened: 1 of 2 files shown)


michaelsauter:
FYI @segator @borja44 This will affect the docs, right? They list this param? Maybe it would be good to check, after this PR is merged, if the docs reflect the actual param (names).

segator:
Maybe anyway change this is not a problem :)

@borja44 can you double confirm?

michaelsauter:
@georgfedermann I tried to figure out how this param is populated for the EDP box. But it does not seem to be set in `create_configuration()` - do you know more about it?

georgfedermann:
@michaelsauter actually, it gets set in create_configuration() as a side effect of setting PIPELINE_TRIGGER_SECRET.
Actually a bug that did not cause any effect and went undiscovered so far.

PR #731, opened 2020-07-28
Feature/54 improve install doc

### LINKED ISSUE #730: Improve install documentation
I recently used the [install documentation](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html) to install (a second) OpenDevStack on a OpenShift cluster. During this I faced a few, mostly minor issues or doc gaps that could be addressed to make things clearer. 

Thus, here are my suggestions. Some seem like a no-brainer but since you're just going through a list for this installation you might not think everything through...:

[Local Setup section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_local_setup)
- It was unclear which bitbucket user should be preferably used when running repos.sh... the jenkins cd user or the own account? Doesn't matter?

[Configuration section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_configuration)
- It should be explained that a ods-configuration repo is created locally in which the ods-core.env file is stored, and that this is the file where config params need to be filled into.
- It could also be mentioned, that before commiting, you have to add all files of ods-configuration to git tracking (git add...). 

[Central ODS project section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_central_ods_project)
- An error will tell you, but it could be mentioned to log into oc first before calling the make command.
- You need to have cluster-admin rights on the openshift cluster to perform the installation

[Nexus section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_nexus)
- It is not mentioned that I have to set a new admin password during the make installation nor if this is saved anywhere automatically by the script.

[Sonarqube section](https://www.opendevstack.org/ods-documentation/opendevstack/3.x/administration/installation.html#_sonarqube)
- This part seems outdated or wrong: 
'_The script will prompt for a new admin password, set it and create an auth token to be used by the Jenkins pipelines. Both values can be written into ods-configuration/ods-core.env, which you then need to commit and push before continuing._' 
When running the script I had to enter a new password and the auth token was set automatically. Also, both were automatically written into ods-core.env IIRC.
- Before pushing the ods-core.env, you need to add the bitbucket repo as remote origin, otherwise you can't push... This could also already be done in the Local Setup section I guess.
- You have to add the cd user as a user in the project permissions of your bitbucket ods project. Read permission is sufficient.

Please comment for changes or additions. Thanks.

renedupont:
fixes #730

### DIFF
--- a/docs/modules/administration/pages/installation.adoc
+++ b/docs/modules/administration/pages/installation.adoc
@@ -147,7 +147,7 @@ Afterwards, run the initial configuration:
 make configure-nexus
 ----
 
-In the course of this you have to set a new admin password.
+In the course of this you will be prompted to set a new admin password.
 
 WARNING: The `configure-nexus` target is not idempotent yet, so it cannot be used for upgrading!
 
--- a/docs/modules/administration/pages/installation.adoc
+++ b/docs/modules/administration/pages/installation.adoc
@@ -52,7 +52,11 @@ make prepare-config
 
 If you want to install OpenDevStack for the first time, leave the BitBucket URL empty.
 
-After the configuration is prepared (which means that a directory `ods-configuration` will be created if it does not exist yet), fill out all the parameters in `ods-core.env` in directory `ods-configuration` for your installation. Add all files of directory `ods-configuration` to git and commit the result. You don't need to push yet, this will be done later.
+After the configuration is prepared (which means that a directory `ods-configuration` will be created if it does not exist yet),
+fill out all the parameters in `ods-core.env` in directory `ods-configuration` for your installation.
+Add all files of directory `ods-configuration` to git and commit the result. You don't need to push yet, this will be done later.
+
+"Preparing" also updates the sample files and checks if any params are present in the sample files but missing from the real file.
 
 === Tailor
 
@@ -114,6 +118,7 @@ make sync-repos
 You also need to update `ods-configuration`, which can be done via:
 [source,sh]
 ----
+git remote add origin <bitbucket-ods-configuration-git-repo-url>
 git push -u origin master
 ----
 
@@ -162,8 +167,8 @@ make install-sonarqube
 ----
 
 This will launch an instance of SonarQube.
-The script will prompt for a new admin password, once set, an auth token that is used by the Jenkins pipelines will be generated.
-Both values are automatically written into `ods-configuration/ods-core.env`, which you then need to commit and push after adding the Bitbucket `ods-configuration` repo as the remote origin.
+The script will prompt for a new admin password. Once this has been set, an auth token for the Jenkins pipelines will be generated.
+Both values are automatically written into `ods-configuration/ods-core.env`, which you then need to commit and push.
 
 Also, you have to add the cd-user as a user in the project permissions of your Bitbucket OpenDevStack project. Read permission is sufficient.
 
--- a/docs/modules/administration/pages/installation.adoc
+++ b/docs/modules/administration/pages/installation.adoc
@@ -34,7 +34,7 @@ If you are installing OpenDevStack for the first time, run:
 ./repos.sh --init
 ----
 
-If there is already an existing installation of OpenDevStack, run:
+If there is already an existing installation of OpenDevStack, run the following with a user that has write permission:
 [source,sh]
 ----
 ./repos.sh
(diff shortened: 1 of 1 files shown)


docs/modules/administration/pages/installation.adoc
```diff
@@ -34,7 +34,7 @@ If you are installing OpenDevStack for the first time, run:
 ./repos.sh --init
 ----
 
-If there is already an existing installation of OpenDevStack, run:
+If there is already an existing installation of OpenDevStack, run the following with a user that has write permission:
```
michaelsauter, line 37: Why only when there is already an existing installation?

docs/modules/administration/pages/installation.adoc
```diff
@@ -52,7 +52,7 @@ make prepare-config
 
 If you want to install OpenDevStack for the first time, leave the BitBucket URL empty.
 
-After the configuration is prepared (which means that a directory `ods-configuration` will be created if it does not exist yet), fill out all the parameters for your installation and commit the result. You don't need to push yet, this will be done later.
+After the configuration is prepared (which means that a directory `ods-configuration` will be created if it does not exist yet), fill out all the parameters in `ods-core.env` in directory `ods-configuration` for your installation. Add all files of directory `ods-configuration` to git and commit the result. You don't need to push yet, this will be done later.
```
michaelsauter, line 55: "Preparing" also updates the sample files and checks if any params are present in the sample files but missing from the real file.

docs/modules/administration/pages/installation.adoc
```diff
@@ -159,8 +162,10 @@ make install-sonarqube
 ----
 
 This will launch an instance of SonarQube.
-The script will prompt for a new admin password, set it and create an auth token to be used by the Jenkins pipelines.
-Both values can be written into `ods-configuration/ods-core.env`, which you then need to commit and push before continuing.
+The script will prompt for a new admin password, once set, an auth token that is used by the Jenkins pipelines will be generated.
```
michaelsauter, line 165: I think there is a typo in this sentence? Not sure how to read/understand it ...

docs/modules/administration/pages/installation.adoc
```diff
@@ -159,8 +162,10 @@ make install-sonarqube
...
+Both values are automatically written into `ods-configuration/ods-core.env`, which you then need to commit and push after adding the Bitbucket `ods-configuration` repo as the remote origin.
```
michaelsauter, line 166: Isn't the user asked earlier in the guide to push this repo already? Then the remote should be set already.

docs/modules/administration/pages/installation.adoc
```diff
@@ -159,8 +162,10 @@ make install-sonarqube
...
+
+Also, you have to add the cd-user as a user in the project permissions of your Bitbucket OpenDevStack project. Read permission is sufficient.
```
michaelsauter, line 168: This should be closer to the `prepare-bitbucket-repos` target, no?

docs/modules/administration/pages/installation.adoc
renedupont: I looked in the script and here https://github.com/opendevstack/ods-core/blob/master/ods-setup/repos.sh#L213 is a push if init = false. If init = true then there seems to be no writing git cmd.

docs/modules/administration/pages/installation.adoc
renedupont: will add it

docs/modules/administration/pages/installation.adoc
michaelsauter: Uh good point. Maybe it shouldn't do that - maybe that should only happen when `--sync` is given.

docs/modules/administration/pages/installation.adoc
renedupont: altered it!

docs/modules/administration/pages/installation.adoc
renedupont: moved it to the bitbucket part!

docs/modules/administration/pages/installation.adoc
renedupont: removed the remote origin part of the sentence!

PR #743, opened 2020-08-03
(Tests) respect bitbucket project during tests

clemensutschig:


tests/utils/clean-up.go
```diff
@@ -75,6 +75,7 @@ func RemoveAllTestOCProjects() error {
 	if err != nil {
 		return err
 	}
+	time.Sleep(20 * time.Second)
```
hugowschneider, line 78: Usually, sleep could create silent problems if for any reason the time needed by the process increases. But if there is no other way

tests/utils/clean-up.go
```diff
@@ -93,6 +94,7 @@ func RemoveAllOpenshiftNamespacesForProject(project string) error {
...
```
hugowschneider, line 97: Same here

PR #765, opened 2020-08-10
add 20mi min storage - for prov app pvc

clemensutschig:
fix for: 

encountered error(s) in creating the volume: [Failed to create volume odstest-prov-history-cc1e0 on storage pool nahsafascl03_01_data01__fileservice from backend ontapnas_nahsafascl03-cs01-nfs-01: requested volume size (10485760 bytes) is too small; the minimum volume size is 20971520 bytes]


clemensutschig:
@borja44  -f yi

PR #777, opened 2020-08-21
Add framework to run quickstarter tests

michaelsauter:
This replaces the test setup in ods-quickstarters - it basically moves the Go code from there to this repo.

Instead of having the test logic in the quickstarter repository to test, we
define the tests once in ods-core, and all quickstarters can be tested
using that framework by just bringing a "testdata" directory which
contains the steps to execute, and the expected golden files.

This has several advantages:
* It makes it easier to test quickstarters not located in
ods-quickstarters
* It avoids to duplicate code (e.g. shell scripts) between ods-core and
ods-quickstarters
* It makes the dependency obvious - previously one needed to run the
tests in ods-core first in order to run the ones in ods-quickstarters.
This is now more obvious and can be enforced by tooling.

Apart from this, the PR addresses all complaints by the linter (and
runs linting during CI now), and generally improves the Go code as well as the
bash scripts. Note though that the added bash scripts are just copied
from ods-quickstarters and not touched for the most part to keep impact
low.

In terms of impact, this PR is `+1,577 -164` but the companion PR in ods-quickstarters is `+475 −3,622` - which means a total of `-1,734` lines.


michaelsauter:
@clemensutschig @georgfedermann I made adjustments to `deploy.sh` now, which should do the trick.

michaelsauter:
@clemensutschig @georgfedermann This is ready to be merged now. OK from your side?

clemensutschig (approved):

PR #786, opened 2020-08-27
ods-core.env in LF endling mode

segator:
ods-core.env in CRLLF endling mode causes issues when installation or executing ods-core tests
Some issues found:

* Invalid URL's
* base64 can not decode/undecode
* Invalid passwords
* Invalid users

### DIFF
--- a/docs/modules/administration/pages/installation.adoc
+++ b/docs/modules/administration/pages/installation.adoc
@@ -58,6 +58,8 @@ Add all files of directory `ods-configuration` to git and commit the result. You
 
 "Preparing" also updates the sample files and checks if any params are present in the sample files but missing from the real file.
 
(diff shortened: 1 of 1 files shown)


docs/modules/administration/pages/installation.adoc
```diff
@@ -58,6 +58,8 @@ Add all files of directory `ods-configuration` to git and commit the result. You
 
 "Preparing" also updates the sample files and checks if any params are present in the sample files but missing from the real file.
 
+WARNING: If you are using windows **Cygwin** or **WSL** remember to change ending line CRLF to LF before commit `ods-core.env`
```
michaelsauter, line 61: How about linking to https://docs.github.com/en/github/using-git/configuring-git-to-handle-line-endings? There's a "Windows" section which explains how to apply some global config to avoid issues.

segator: good idea 👍

michaelsauter (approved):

PR #794, opened 2020-09-02
Add OpenVPN setup to ods box

georgfedermann:


michaelsauter:
Cool - could you please add some explanation on why  / what this is for?

georgfedermann:
This change is required to support a local e.g. prov-app development environment where services to integrate with (like Atlassian Stack) are provided by an ODS box instance. The client machine running the development environment will be connected to the ODS box via VPN and will be able to resolve services by their names as configured in the ODS box's DNS service.

clemensutschig (approved):

Issue #817, opened 2020-09-22
Parameterization of CPU req/limits in OpenShift templates


felipecruz91:
**Is your feature request related to a problem? Please describe.**
When carrying out a new ODS installation, we may find the necessity to configure how much CPU and memory request/limits we want to establish for Sonar, Nexus, Prov app and Jenkins. As of version 3.x, those values seem to be hard-coded in the Tailor Template yml files and cannot be overriden.

**Describe the solution you'd like**
An option could be to extend the Tailor .yml files so that it reads the parameters from variables that could be set in the `ods-core.env` file, for instance. If no variable has been found in `ods-core.env` or other `.env` file, Tailor will use the hard-coded values in its Template yml file.

**Example**
For instance, if I wanted to increase the Nexus memory limit constraint (which is currently [hard-coded](https://github.com/opendevstack/ods-core/blob/master/nexus/ocp-config/bc.yml#L21) to 2Gi), I could define that variable in `ods-core.env` and extend the [bc.yml](https://github.com/opendevstack/ods-core/blob/master/nexus/ocp-config/bc.yml) file to handle this scenario.

Any input would be very much appreciated, thanks 👍

michaelsauter:
The templates are already parameterised: https://github.com/opendevstack/ods-core/blob/master/nexus/ocp-config/bc.yml#L14-L21.

What you see there are just default values.

If you set one of those parameters in your `ods-core.env` file, they'll get used :)

So maybe this just needs to be documented?

felipecruz91:
That's really great news then!

If you could point me to the right section where you'd like this to be documented, I'd be happy to do it.

michaelsauter:
Not sure. Maybe as comments in `ods-core.env` that point to the templates for more customisation? Where would you have searched for / found this information?

felipecruz91:
Yes, `ods-core.env` is a good starting point to document it.

michaelsauter:
I think this is done, right? I'll close the issue. If there are still some params missing, please open a new issue for them.

PR #818, opened 2020-09-23
Update ods-core.env.sample

felipecruz91:
Added Nexus, SonarQube and ProvApp resource request/limit link that points to the templates - (Issue #817)

[Jenkins](https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/build/bc.yml#L50-L56) and [DocGen](https://github.com/opendevstack/ods-core/blob/master/ods-document-generation-svc/ocp-config/bc.yml#L36-L42) resources are hard-coded unfortunately.


configuration-sample/ods-core.env.sample
```diff
@@ -46,6 +46,10 @@ NEXUS_PASSWORD_B64=changeme
 # Nexus (basic) authentication username:password
 NEXUS_AUTH=developer:changeme
 
+# Nexus request/limits.
+# You can configure them here: https://github.com/opendevstack/ods-core/blob/master/nexus/ocp-config/bc.yml#L14-L21
+# For instance,
+# NEXUS_BUILD_CPU_REQUEST=200m
```
michaelsauter, line 52: For me this sounds like people should edit https://github.com/opendevstack/ods-core/blob/master/nexus/ocp-config/bc.yml#L14-L21. But that's not really what we want ... I think we should tell the user that all params in that template can be overwritten by specifying the param in this file here ...

michaelsauter:
@felipecruz91 Do you want to make the hardcoded instance parameterised as well? Would be good for consistency ..

configuration-sample/ods-core.env.sample
felipecruz91: That's right - it gives a false impression that they should edit the params, well spotted!

felipecruz91:
> @felipecruz91 Do you want to make the hardcoded instance parameterised as well? Would be good for consistency ..

@michaelsauter Sure! I'll be creating 2 separate PRs, one for Jenkins and another one for Docgen.

michaelsauter (approved):

PR #828, opened 2020-09-30
add jenkins smtp configuration docs

### LINKED ISSUE #827: configure jenkins email server configs
Current Jenkins master deployments do not come with defaults and/or configurable email server settings. Users need to configure it manually.
Default configuration expects a localhost SMTP relay/server working, which is usually not the case.

Would be nice to have some or all of the following requests:
- provide a default configuration from scratch (if possible) that enables the email notifications from an ODS point of view
- provide a configurable way via lib context (dynamic) or env vars (not that dynamic); or both?
  --> I think that the unique way to handle it will be for now manual user interaction with the Jenkins configuration: https://www.jenkins.io/doc/pipeline/steps/email-ext/
- document it/note it in the ODS docs

The `SMTP server` param configuration that needs to be defined is found under menu option:
`Jenkins` -> `Manage Jenkins` -> `Configure System` -> `Extended E-mail Notification`

To avoid ending up in spam folder, one might also want to be able to provide a known origin FROM default email param value. The param is called `System Admin e-mail address` and this is defined under menu option: `Jenkins` -> `Manage Jenkins` -> `Configure System` -> `Jenkins Location`

**Describe alternatives you've considered**
To at least have this documented in the ODS docs

**Additional context**
More information on what is used underneath:
- https://plugins.jenkins.io/mailer/
- https://javaee.github.io/javamail/#overview_description

gerardcl:
fixes #827

### DIFF
--- a/docs/modules/jenkins/pages/master.adoc
+++ b/docs/modules/jenkins/pages/master.adoc
@@ -30,6 +30,27 @@ specifies the location of the SonarQube server.
 SONAR_AUTH_TOKEN::
 this is the SonarQube auth token that is generated when deploying and configuring sonarqube
 
-== Mail Sending
+== Email notifications
 
-In the Dockerfile, sending of mails to unknown users and users without read rights is enabled.
+Jenkins email notifications need manual configuration:
+
+1. The `SMTP server parameters` that need to be defined are found under menu option:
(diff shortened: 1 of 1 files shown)


metmajer (approved):
Many thanks @gerardcl!

michaelsauter (approved):

Issue #773, opened 2020-08-13
Adjust versioning slightly


michaelsauter:
The current versioning scheme is documented at https://www.opendevstack.org/ods-documentation/opendevstack/3.x/getting-started/index.html#_versioning.

I want to propose the following adjustments:

* We notice that there is a need for more frequent "feature releases" which are backwards compatible. People typically talk about them as "3.1", "3.2" etc. However, there is also a need to do bugfix releases, and they should be distinguishable from a "feature release". Therefore we should document that Git tags have 3 digits: MAJOR.MINOR.PATCH, and that patch releases are possible. Note that we already tagged ODS 3 with 3 digits - `v3.0.0`.

* Currently users install ODS most likely with `ODS_GIT_REF=3.x`, which means they point the installation against a branch, and not a tag. This is required because the value of `ODS_GIT_REF` is used when the `Jenkinsfile.template` of a quickstarter is rendered (as the pointer to the Jenkins shared library), and we don't want to require users to update their Jenkinsfiles every time there is a compatible (patch) release. However, it would be better (e.g. for documentation purposes) if installations are based on a tag, to make it crystal clear which sources are used. Therefore, I propose change one small detail: to use the value of `ODS_IMAGE_TAG` instead of `ODS_GIT_REF` when rendering the Jenkins shared library reference  `Jenkinsfile.template`. We then need to ensure during installation that the value of `ODS_IMAGE_TAG` exists as a Git reference in `ods-jenkins-shared-library`, and that it points to the same Git commit as the Git reference specified in `ODS_GIT_REF`.

We could also think about renaming `ODS_IMAGE_TAG` to `ODS_VERSION` so that it isn't confusing when that value is used for the Jenkins shared lib ref. But honestly not sure if this minor detail is worth the effort. One technical note which is also relevant for this decision: during testing of quickstarters, it has proven beneficial to be able to specify the Jenkins shared library and agent image tags separately, so I would still have different input variables. E.g. `env.AGENT_IMAGE_TAG` and `env.JENKINS_SHARED_LIBRARY_REF`, and just fill them, if not set, with `env.ODS_IMAGE_TAG`

@metmajer @clemensutschig @martsec @oalyman @henrjk @stitakis @gerardcl Would be good to hear your thoughts on this. I don't see other exceptions than the one "jenkins shared library reference". Maybe you do see one? Or you have alternative suggestions / ideas?

I would propose to make this change for ODS 4, making it the first version wich can and should be installed from a Git tag.

[...]

PR #831, opened 2020-10-08
Shared library git ref

michaelsauter:
Addresses #773.

The main goal is to be able to use a Git tag for `ODS_GIT_REF` to allow admins to nail down exactly what version of ODS they install.

To achieve this, the PR also includes the following required changes:

* For users, we only want to communicate the "ODS version" as configured by `ODS_IMAGE_TAG` (which can be as precise or relaxed as admins want, e.g. it can be `4.x`, `4.0.x` or `4.0.0`). Therefore, this PR creates a `make` target that ensures a Git ref equal to `ODS_IMAGE_TAG` exists in `ods-jenkins-shared-library` so that this reference can be consumed by users from their `Jenkinsfile`. To finalise this particular change, we need another PR in the shared library to expose the value of `ODS_IMAGE_TAG` as `agentImageTag` and `sharedLibraryRef` to quickstarters (https://github.com/opendevstack/ods-jenkins-shared-library/pull/457), and finally a PR in the quickstarters to use `agentImageTag` and `sharedLibraryRef`
* Adjust the `ods-setup/repos.sh` script to handle Git tags properly. The PR achieves this by reworking `ods-setup/repos.sh`, which has been split into two scripts. The first script, `scripts/repos.sh` is supposed to be used from the umbrella directory to initialise and update local repos (from GitHub). The script is now safe to be used for initial installation and update using the exact same parameters. The second script, `scripts/push-local-repos.sh`, can be used to push those local repos to Bitbucket. This split makes each script easier to understand, both for ODS developers but also from a user perspective (e.g. because there are fewer parameters now)
* Adjust the ods-devenv box creation to make use of the new scripts and set the shared lib ref


clemensutschig:
@michaelsauter  - cool, but this has very likely major impact also on ods-devenv / deploy.sh or?

michaelsauter:
@clemensutschig Yes this affects ods-devenv, though only slightly. I adjusted the `deploy.sh` script (and also the corresponding installation docs), however I have not tested this part yet. @georgfedermann is this easy for you to run end-to-end as an AMI build? At this stage, it should be "just the same as before" and when we merge the PR in the shared lib / the quickstarters, then we are able to use a Git tag as the ODS source, and have a "virtual identifier" such as `4.x` as the image tag / shared lib ref for users to consume.

michaelsauter:
Marked this PR ready for review now. I adjusted the description above to clearly state the goal and the included changes.

@metmajer @henrjk @segator Pinging you here if you also have comments / concerns / feedback regarding this.

scripts/repos.sh
```diff
@@ -0,0 +1,160 @@
...
+    echo "  GIT_REF=master"
+    echo "  mkdir -p \$UMBRELLA_DIR && cd \$UMBRELLA_DIR"
+    echo "  curl -LO https://raw.githubusercontent.com/opendevstack/ods-core/\$GIT_REF/scripts/repos.sh"
+    echo "  chmod +x repos.sh"
+    echo "  ./repos.sh --git-ref \$GIT_REF"
+    echo ""
+    exit 1
+fi
+
+echo_done(){
```
henrjk, line 30: Could add a comment that since this is used for bootstrapping it is not using colored_output.sh

scripts/repos.sh
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env bash
```
henrjk, line 1: The script was moved, but in the PR I see it as delete and add which makes it harder to compare. Is it worthwhile to diff and review in detail?

henrjk (commented):
This looks good to me and I like the much clearer installation adocs.
If needed I can take a more detailed look at `repos.sh`(see comment about this)

scripts/repos.sh
michaelsauter: Added as a comment in 10d7e94.

scripts/repos.sh
michaelsauter: It was moved, yes, but it has also been slimmed down considerably. Previously it was potentially also pushing (syncing) to Bitbucket, whereas now it is just clone/update from a remote. The general approach hasn't been changed though so I think you don't need to review in detail.

michaelsauter:
@henrjk Thanks for loving at this! I addressed your comments. Could you approve?

@segator @felipecruz91 Talking this over with you now, could you also approve from your side?

@georgfedermann Would you be able to check the part that touches the AMI build? That would be great.

clemensutschig:
@georgfedermann - merge it into ods-devenv

Michael Sauter <notifications@github.com> schrieb am Mo., 26. Okt. 2020,
17:07:

felipecruz91 (approved):

henrjk (dismissed):

renedupont (dismissed):

clemensutschig:
@michaelsauter  - did you merge this into ods-devenv?

michaelsauter:
@clemensutschig No, I did not hear back from @georgfedermann yet. Anyway, I decided it is finally time to get some first-hand experience building this AMI so I'm setting up access to an AWS account at the moment and then I'll check for myself :)

georgfedermann:
ah, I thought I should also merge it into feature/ods-devenv as soon as it's approved. I did not know that it was about testing the change. Sorry about that. @michaelsauter can we create a PR against feature/ods-devenv?

georgfedermann:
sorry for the delay, this will be on the midnight build for feature/ods-devenv branch.

clemensutschig:
@michaelsauter  - looks like ods-devenv broke / can you work with @georgfedermann  to see whats going on?

georgfedermann:
@michaelsauter @clemensutschig the current build breaking issue was that packer would execute an outdated bootstrap.sh script. Sorry bout that, I fixed it for the 18:00 build.

georgfedermann:
@michaelsauter
After the fix of the packer scripts, the build for feature/ods-devenv is still failing.
After line repos.sh:125 the branch GIT_REF will not be checked-out / available locally for repository ods-document-generation-templates, and thus the build will break in push-local-repos.sh:100, when it tries to push feature/ods-devenv to bitbucket.

michaelsauter:
@georgfedermann Thanks for the detailed explanation. I missed to move the special handling of `ods-document-generation-templates` to `push-local-repos.sh`. Fixed in e329fb3.

georgfedermann:
@michaelsauter 
works! Please merge

PR #833, opened 2020-10-08
exposes odsbox jira jdwp port

stitakis:
Currently odsbox jira instance do not expose the java debug port for remote debugging. 
This change adds the docker container parameter to the jira instance to enable and expose the jdwp.


georgfedermann (approved):
💯

PR #834, opened 2020-10-08
WIP: add to provisioning app a service account token as env var

stitakis:
partially fix ProvApp issue https://github.com/opendevstack/ods-provisioning-app/issues/589

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,9 @@ start-provisioning-app-build:
 
 ## Update OpenShift resources related to the Provisioning App service.
 apply-provisioning-app-deploy:
-	cd ods-provisioning-app/ocp-config && tailor apply --namespace $(ODS_NAMESPACE) --exclude is,bc
+	# first get from openshift the name of the default service account token which is used by prov app to interact with oc api
+	SA_TOKEN=$$(oc project $(ODS_NAMESPACE) | oc get sa -o=json | jq -r '.items[] | select(.metadata.name=="default").secrets[] | select(.name|test("^(default-token).+")).name') ;\
+ 	cd ods-provisioning-app/ocp-config && tailor apply --namespace $(ODS_NAMESPACE) --param ODS_DEFAULT_SA_TOKEN=$$SA_TOKEN --exclude is,bc
 .PHONY: apply-provisioning-app-deploy
 
 # DOCUMENT GENERATION SERVICE IMAGE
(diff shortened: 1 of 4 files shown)


michaelsauter:
I'm not sure about the direction here.

To my knowledge, the prov app does not interact with OpenShift directly. This change opens up this can. Before we proceed, can we first define what problem we are solving and discuss if this is the way we want to go? Would appreciate this.

stitakis:
@michaelsauter sure, np... the problem is defined in https://github.com/opendevstack/ods-provisioning-app/issues/589... the ProvApp should be able to do a pre flight check to verifies that the project name does not exists in openshift before creating it. To be able to do that the ProvApp needs a way to authenticate on the openshift api in order to get a list of existing project.
Is this direction ok, or should we go in a different direction? If is not ok, in which direction should we go to enable ProvApp to be able to query from openshift api the list of existing projects?

michaelsauter:
Hmm.

opendevstack/ods-provisioning-app#589 indicates that the problem arises from the E2E tests? While I see the gap in the checks, it would be great if we could avoid adding lots of functionality just to prevent some test failures that we could potentially solve in an easier fashion.

So ... putting out two ideas ... I don't have others so far. Do any of those help?

* Do we really need to stop? Or could the Jenkins job just not do anything if the project already exists? That could have the advantage that you can run the same thing over and over again (e.g. if some failures occurred) and only the missing pieces are created? I haven't thought this through yet, but could be worthwhile to evaluate.
* If we use the serviceaccount, then we can simplify the implementation: instead of reading and injecting the token, we can just either use a token given by an env var, and if that is not present, read the token from the file in the pod (see https://github.com/opendevstack/ods-core/blob/master/jenkins/webhook-proxy/main.go#L25). That way the user does not need to change the config, and this PR is not needed at all - you only need to do changes in the prov app repo.

stitakis:
@michaelsauter thanks for the feedback... I'll come back on this after my holiday next week

stitakis:
@michaelsauter I was able to take a look at this....

the checks are not only need for E2E tests. This is not expressed in the issue page above. I'll update it.
Main goal of the pre flight checks is to guarantee that the project provision will not fail. Without them if the project provision fails, it ends up leaving an inconsistent state which requires manual cleanup changes in different servers.

Yes, the suggested simplified implementation of reading the service account from the path above works fine for ProvApp too. I will follow this approach. So this PR is not needed anymore.

Thanks!

PR #848, opened 2020-10-20
Set better default for ALLOWED_EXTERNAL_PROJECTS

michaelsauter:
This prevents having to set the env var `ALLOWED_EXTERNAL_PROJECTS` on existing webhook proxy instances when users update from 2.x to 3.x, and admins changed the default Bitbucket project from `opendevstack` to something else.

FYI @segator @felipecruz91 

I'll backport to 3.x once approved/merged.


segator (approved):
Thanks!

PR #853, opened 2020-10-27
Add volumeMode field to PVC resources

michaelsauter:
This field has been added in Kubernetes 1.18, see
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#volume-mode.

OpenShift 3.11 uses an older Kubernetes version, which ignores the field
if specified.

OpenShift 4 uses Kubernetes 1.18 and therefore defaults the field to
"Filesystem" if not specified, resulting in Tailor to see drift between
desired and current state if the template does not contain the field.

To allow the same template to be used against both OpenShift 3.11 and 4, we
now specify the field in the template, but preserve the live value.

### DIFF
--- a/jenkins/ocp-config/deploy/Tailorfile
+++ b/jenkins/ocp-config/deploy/Tailorfile
@@ -4,5 +4,7 @@ ignore-unknown-parameters true
 
 selector template=ods-jenkins-template
 
+preserve pvc:/spec/volumeMode
+
 svc,route,secret,pvc,dc,rolebinding,serviceaccount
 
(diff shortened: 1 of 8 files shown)


gerardcl (approved):
🚀 LGTM!

gerardcl:
this might not be a breaking change (since PVCs might not be touched at all on running clusters...), but we should document it later on (upgrade guides)!

michaelsauter:
@gerardcl Thanks for looking at this. I don't really understand why you think this should be mentioned more prominently than "just" in the changelog though. Could you explain this? This has no effect on existing installations at all, at least none that I know :)

gerardcl:
I retire my comment, fully agree on only mentioning in changelog, my mind was still (in this regard) at least two versions of tailor in the past, you know.. :)

PR #857, opened 2020-10-28
Urlencode SonarQube parameters

### LINKED ISSUE #798: make configure-sonarqube fails when using special characters
when you use special characters in your password  for example & the target is not skipping the password and make the script fails.

michaelsauter:
The API of SonarQube uses query parameters, which need to be urlencoded.

Fixes #798.

### DIFF
--- a/sonarqube/configure.sh
+++ b/sonarqube/configure.sh
@@ -21,6 +21,12 @@ echo_info(){
     echo -e "\033[94mINFO\033[39m: $1"
 }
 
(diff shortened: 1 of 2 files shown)


clemensutschig (approved):

clemensutschig:
@michaelsauter  - worth to backport to 3.1?

michaelsauter:
@clemensutschig Yes will do!

PR #859, opened 2020-10-29
Use new build API version in webhook proxy

michaelsauter:
The new build API version is compatible with OpenShift 4. It also works
on an OpenShift 3.11 cluster, so using the new path / version we can
support both OpenShift versions.


gerardcl (approved):
LGTM!

PR #862, opened 2020-11-02
Replace script referenced in Jenkins Master entrypoint

michaelsauter:
Since OpenShift 4.3, the Jenkins master image uses an ENTRYPOINT
statement that references a script. Therefore we cannot use CMD anymore
to execute our customized script first.

See
https://github.com/openshift/jenkins/blob/release-4.3/2/Dockerfile.rhel7.

Also, `/opt/openshift/configuration/jobs/OpenShift*` does not seem to exist on UBI based images.

### DIFF
--- a/jenkins/master/Dockerfile
+++ b/jenkins/master/Dockerfile
@@ -20,10 +20,11 @@ RUN import_certs.sh
 COPY plugins.txt /opt/openshift/configuration/plugins.txt
 COPY kube-slave-common.sh /usr/local/bin/kube-slave-common.sh
 RUN /usr/local/bin/install-plugins.sh /opt/openshift/configuration/plugins.txt \
-    && rm -r /opt/openshift/configuration/jobs/OpenShift* \
-    && touch /var/lib/jenkins/configured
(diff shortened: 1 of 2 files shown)


clemensutschig (approved):
@michaelsauter  - is this also working with 3.x OCP

jenkins/master/Dockerfile
```diff
@@ -20,10 +20,11 @@ RUN import_certs.sh
 COPY plugins.txt /opt/openshift/configuration/plugins.txt
 COPY kube-slave-common.sh /usr/local/bin/kube-slave-common.sh
```
clemensutschig, line 21: @michaelsauter  . is there a reason for having this actually still in? ... just asking :)

michaelsauter: Probably not. I created a separate issue for this: https://github.com/opendevstack/ods-core/issues/863

michaelsauter:
@clemensutschig Yes it also works under 3.11. We could have done the customisation like this from the beginning ...

PR #865, opened 2020-11-02
Improve devenv script and docs

michaelsauter:


ods-devenv/scripts/run-on-aws.sh
```diff
@@ -91,45 +100,54 @@ if ! which jq &> /dev/null; then
...
+                --owners ${ami_owner} \
+                --filters "Name=name,Values=CentOS*" \
                 --query 'Images[*].{ImageId:ImageId,CreationDate:CreationDate}' | jq -r '. |= sort_by(.CreationDate) | reverse[0] | .ImageId')
+            if [[ -z "${ami_id}" ]] || [[ "null" == "${ami_id}" ]]; then
+                echo "Did not find vanilla CentOS AMI"
+                exit 1
+            fi
             ec2_instance_name="ODS in a box Install (${target_git_ref}) $(date '+%Y-%m-%d %H:%M:%S')"
-            echo "You are in install mode using CentOS 7 image ${ami_id}."
+            echo "You are in install mode using a vanilla CentOS 7 image ${ami_id}."
```
georgfedermann, line 133: it's actually a custom CentOS 7 2003 AMI for the ODS Box

ods-devenv/scripts/run-on-aws.sh
```diff
@@ -91,45 +100,54 @@ if ! which jq &> /dev/null; then
...
+      echo "Neither --security-group-id nor --iam-instance-profile specified. Using default SG: ${default_securitygroup}"
+      security_group_id="${default_securitygroup}"
     fi
     if [[ -z "${ami_id}" ]]; then
-        if [[ -n "${install}" ]]
-        then
-            ami_id=$(aws ec2 describe-images \
-                --owners 275438041116 \
-                --filters "Name=name,Values=import-ami-*" "Name=root-device-type,Values=ebs" "Name=tag:Name,Values=CentOS*" \
+        if [[ -n "${install}" ]]; then
```
georgfedermann, line 123: btw: I think I'll remove this use case next to install an ODS Box on a CentOS box. We have the packer build for that which will create ODS Box AMIs that can be easily used und quickly instantiated, and run-on-aws.sh is in fact only used to instantiate ODS box instances from ODS Box AMIs. Or do you have different thoughts?

georgfedermann (approved):
lgtm

ods-devenv/scripts/run-on-aws.sh
michaelsauter: Ah! Now I get your comment in our call earlier fully :) I think I agree with the direction - it's definitely better to have just one way to do things, and I like the separation of build vs. run.

michaelsauter:
Given your above comment, let's not merge this as is. I'll look into the packer thing first and then provide an updated PR.

ods-devenv/scripts/run-on-aws.sh
serverhorror: #988 :(

metmajer:
@clemensutschig we are currently investigating the future of ODS in a Box. Code ready containers as a replacement for "oc cluster up" requires nested virtualisation which does not work on AWS unless using bare metal instances.

stitakis:
@clemensutschig @metmajer I wasn't aware of this PR... as main maintainer of the ods box I'll discuss with @michaelsauter and take an action on this

stitakis:
I will close this PR as is pretty old and I'm not sure if it break the ODS build.

PR #867, opened 2020-11-05
Document update procedure of Nexus and SonarQube

michaelsauter:
To make it very easy to ensure a correct local state, repos.sh learns to
deal with the ods-configuration repository.

### DIFF
--- a/docs/modules/nexus/pages/administration.adoc
+++ b/docs/modules/nexus/pages/administration.adoc
@@ -1,0 +1,45 @@
+= Nexus Administration
+
+The `ods-core` repository contains a build recipe for a central Nexus instance. The remainder of this document describes how that setup is done, and how it can be modified and updated.
+
+== Updating Nexus to a new version
+
+The Nexus deployment is using an image built in the central ODS namespace. This image is a customisation of the https://hub.docker.com/r/sonatype/nexus3[official Nexus image on DockerHub].
+
+The Nexus version which is used is determined by which image tag from DockerHub is used, and depending on tag, at which time the ODS image is built.
+
+To update Nexus, the following steps need to be taken:
+
+1. Configure new version
+2. Update `BuildConfig`
+3. Create new image
+4. Rollout new deployment
+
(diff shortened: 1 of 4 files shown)


docs/modules/nexus/pages/administration.adoc
```diff
@@ -0,0 +1,45 @@
+= Nexus Administration
+
+The `ods-core` repository contains a build recipe for a central Nexus instance. The remainder of this document describes how that setup is done, and how it can be modified and updated.
+
+== Updating Nexus to a new version
+
+The Nexus deployment is using an image built in the central ODS namespace. This image is a customisation of the https://hub.docker.com/r/sonatype/nexus3[official Nexus image on DockerHub].
```
felipecruz91, line 7: Where can the reader learn more about the customization mentioned?

docs/modules/nexus/pages/administration.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+The Nexus version which is used is determined by which image tag from DockerHub is used, and depending on tag, at which time the ODS image is built.
```
felipecruz91, line 9: Perhaps a link to the Dockerfile indicating which tag is being used at the moment? https://github.com/opendevstack/ods-core/blob/master/nexus/docker/Dockerfile

stitakis (dismissed):

akhilsoman (approved):
Worked well for us in Nexus upgrade. 👍

docs/modules/nexus/pages/administration.adoc
michaelsauter: Addressed in 2c6cb49.

docs/modules/nexus/pages/administration.adoc
michaelsauter: Addressed in 2c6cb49. Note that the tag in the `Dockerfile` does not represent what actually gets built. This is determined by the `BuildConfig`.

PR #875, opened 2020-11-10
Add script to build AMI via SSM

michaelsauter:
Useful for environments behind proxy that cannot establish connections on port 22.

### DIFF
--- a/ods-devenv/packer/build-ods-ami-ssm.sh
+++ b/ods-devenv/packer/build-ods-ami-ssm.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+set -eu
+set -o pipefail
+# build-ods-ami-ssm.sh builds an ODS image using Packer over SSM.
+# It depends on a prebuilt CentOS7 AMI which has SSM installed.
+# To use this in your AWS account, you must have an IAM instance profile setup
+# as described in https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-instance-profile.html#instance-profile-add-permissions.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+iam_instance_profile="SSMInstanceProfile"
+profile="default"
+region="eu-west-1"
+instance_type="m5ad.4xlarge"
+ods_branch="master"
+
+function usage {
+    printf "Builds an ODS image using Packer over SSM.\n\n"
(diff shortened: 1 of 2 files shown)


ods-devenv/packer/build-ods-ami-ssm.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+set -eu
+set -o pipefail
+# build-ods-ami-ssm.sh builds an ODS image using Packer over SSM.
+# It depends on a prebuilt CentOS7 AMI which has SSM installed.
```
stitakis, line 5: This line mentions that this scripts depends on a prebuild CentOS7 with SSM installed, however in the code I can't see why SSM is required. Is there a hidden dependency? If yes, can you please a line explaining this?

michaelsauter: It's in the packer JSON. That is configured to use SSM instead of SSH.

stitakis (approved):

ods-devenv/packer/build-ods-ami-ssm.sh
```diff
@@ -0,0 +1,59 @@
...
+# To use this in your AWS account, you must have an IAM instance profile setup
+# as described in https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-instance-profile.html#instance-profile-add-permissions.
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+iam_instance_profile="SSMInstanceProfile"
```
georgfedermann, line 11: awesome! works like a charme.
Would be great if iam_instance_profile could be configured via arguments, too.

ods-devenv/packer/build-ods-ami-ssm.sh
```diff
@@ -0,0 +1,59 @@
...
+profile="default"
+region="eu-west-1"
+instance_type="m5ad.4xlarge"
+ods_branch="master"
+
```
georgfedermann, line 16: also to have the (private) VPC subnet configurable where the build box should be run.

stitakis:
@michaelsauter do we need this change. Otherwise let us close it.

michaelsauter:
@stitakis I don't know to be honest. I never use the box (except for this one time). It seems like no-one else needed it so let's close it.

Issue #687, opened 2020-07-06
Update Skopeo to latest version


michaelsauter:
We install 0.1.37, but the latest release as of today is 1.1.0: https://github.com/containers/skopeo/releases. Unfortunately, the mirrors we use don't have 1.1.0, and `yum install skopeo` fails to find the package in our base image :(

michaelsauter:
See #876.

PR #876, opened 2020-11-10
Updated skopeo to v1.2.0

felipecruz91:
Refers to #687

### DIFF
--- a/jenkins/agent-base/Dockerfile
+++ b/jenkins/agent-base/Dockerfile
@@ -6,7 +6,7 @@ ENV SONAR_SCANNER_VERSION=3.1.0.1141 \
     CNES_REPORT_VERSION=3.1.0 \
     TAILOR_VERSION=1.2.2 \
     GIT_LFS_VERSION=2.6.1 \
-    SKOPEO_VERSION=0.1.37-3 \
+    SKOPEO_VERSION=1.2.0 \
     OSTREE_VERSION=2018.5-1
 
 ARG APP_DNS
@@ -44,10 +44,10 @@ RUN cd /tmp \
 
 # Install Tailor.
 RUN cd /tmp \
-	&& curl -LOv https://github.com/opendevstack/tailor/releases/download/v${TAILOR_VERSION}/tailor-linux-amd64 \
-	&& mv tailor-linux-amd64 /usr/local/bin/tailor \
-	&& chmod a+x /usr/local/bin/tailor \
-	&& tailor version
+    && curl -LOv https://github.com/opendevstack/tailor/releases/download/v${TAILOR_VERSION}/tailor-linux-amd64 \
+    && mv tailor-linux-amd64 /usr/local/bin/tailor \
+    && chmod a+x /usr/local/bin/tailor \
+    && tailor version
 
 # Install GIT-LFS extension https://git-lfs.github.com/.
 RUN cd /tmp \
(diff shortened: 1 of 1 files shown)


michaelsauter:
Cool!

The tests are failing right now though.

Also, we're in the process to build agent images on UBI (for OpenShift 4) and installing Skopeo is also a problem there. I'll get back to this PR once we have collected a little more experience with UBI8 and how we want to solve the issue there.

clemensutschig:
@michaelsauter  - repo download magic - rerun fixed it 👍

michaelsauter:
#879 uses CentOS 8 repos now for the UBI8 image. Those have an older version of Skopeo right now (0.41), but it doesn't require dealing with yet another repo. Right now I'm not aware of a use-case that requires 1.2.0 instead of 0.41 so I would be ok with the older version.

My proposal is to update the base agent for OpenShift 3.11  to get access to CentOS 7 repos, and then install Skopeo from there.

serverhorror:
> Right now I'm not aware of a use-case that requires Skopeo 1.2.0 instead of 0.41 so I would be ok with the older version.

With the current image it is hard/impossible to log in via docker to registries that require authentication.

* [Private registries with authentication](https://github.com/containers/skopeo/tree/v0.1.41#private-registries-with-authentication)

We are using our own image to have the docker client available so we are able to log in to an ECR registry and push the image. The first mention of skopeo where a re `skopeo login ...` is available is at 1.0.0:

* [Authenticating to a registry](https://github.com/containers/skopeo/tree/v1.0.0)

michaelsauter:
@serverhorror Thanks for pointing this out! However, I'm not sure I already get the issue. Why do you need the `docker` client binary?

I would assume usage like this:
```
skopeo copy \
  docker://source-registry.example.com/foo/bar:latest \
  docker://dest-registry.example.com/foo/bar:latest \
  --src-creds source-user:source-token \
  --dest-creds dest-user:dest-token
```

Above usage works with`0.1.37`. How are you invoking `skopeo`?

serverhorror:
@michaelsauter This is what we currently need to do.

Pay attention `#Login to docker registries` -- I couldn't get skopeo to even read from the "internal" cluster without all of this.

Things to note:

* `ocRegistry` is the *public* endpoint of the OpenShift registry, skopeo didn't like the internal one at all and I couldn't get access to the CA (do note that this is running in a custom image because I needed to get the docker client tools installed)
* `imageRegistry` points to a repository living in AWS ECR

```groovy
sh(label: "Login to docker registries and copy images",
  script: """
    DOCKER_CONFIG=\$(mktemp -d)
    export DOCKER_CONFIG
    export DOCKER_TLS_VERIFY=0

    curl --silent --insecure --location --output "\$DOCKER_CONFIG/bi_root.crt" 'https://pki.boehringer.com/pki/INHASRootCA.boehringer.com_Boehringer%20Root%20CA.crt'
    openssl x509 -in "\$DOCKER_CONFIG/bi_root.crt" -inform DER -out "\$DOCKER_CONFIG/bi_root.pem" -outform PEM

    curl --silent --insecure --location --output "\$DOCKER_CONFIG/bi_issuing.crt" 'https://pki.boehringer.com/pki/inhas01822.boehringer.com_Boehringer%20Issuing%20CA%2001.crt'
    openssl x509 -in "\$DOCKER_CONFIG/bi_issuing.crt" -inform DER -out "\$DOCKER_CONFIG/bi_issuing.pem" -outform PEM

    cat "\$DOCKER_CONFIG/bi_root.pem" > "\$DOCKER_CONFIG/ca.pem"

    # Login to docker registries
    cat /run/secrets/kubernetes.io/serviceaccount/token | docker login --username default --password-stdin "${ocRegistry}"
    aws ecr get-login-password | docker login --username AWS --password-stdin "${awsRegistry}"

    # copy the image from the 'default' registry to the target registry
    skopeo copy --authfile "\$DOCKER_CONFIG/config.json" --src-tls-verify=false \\
      "docker://${ocRegistry}/${context.targetProject}/${context.componentId}${componentSuffix}:${context.shortGitCommit}" \\
      "docker://${imageRepository}:${context.shortGitCommit}"
    skopeo copy --authfile "\$DOCKER_CONFIG/config.json" --src-tls-verify=false \\
      "docker://${ocRegistry}/${context.targetProject}/${context.componentId}${componentSuffix}:${context.shortGitCommit}" \\
      "docker://${imageRepository}:latest"
    echo 'All images copied!'
        """)
```

serverhorror:
The 2 main reasons why I'm doing the docker login dance:

1. go to https://github.com/containers/skopeo#commands and find that it documents the most recent version
1. forget about switching branches to get the right docs
1. `skopeo login` yields error messages
1. frustration and workarounds :)

So, in a sense, PEBCAK. I can definitely live without it but it would be nice if we just get a recent version.

michaelsauter:
I'm going to close this as:

1. new `Dockerfile.ubi8` installs Skopeo from CentOS8 repos
2. if `Dockerfile.rhel7` should be aligned, we should try to install from CentOS7 repos
3. this PR has conflicts now

@felipecruz91 Thanks for opening this though. Let me know if you want to take a shot at (2), but I'd see it as low-pro right now.

PR #879, opened 2020-11-13
Agent Base on UBI8

michaelsauter:
Leaving this as a draft until all tests on top of this agent ran.

Note that this installs Tailor 1.3.0 in the rhel8 variant, while keeping 1.2.2 in the rhel7 variant at the moment. There is no need to update Tailor for OpenShift 3.11, and we do not have automated tests of Tailor against OpenShift 4, so I'm doing the least risky move.


jenkins/agent-base/Dockerfile.rhel7
```diff
@@ -72,11 +72,10 @@ RUN if [ -z $SNYK_DISTRIBUTION_URL ] ; then echo 'Skipping snyk installation!' ;
...
 COPY set_java_proxy.sh /tmp/set_java_proxy.sh
-RUN cp /usr/local/bin/run-jnlp-client /usr/local/bin/ocp-run-jnlp-client
-COPY ods-run-jnlp-client /usr/local/bin/run-jnlp-client
-RUN chmod +x /usr/local/bin/run-jnlp-client \
-    && chmod +x /tmp/set_java_proxy.sh \
-    && . /tmp/set_java_proxy.sh && echo $JAVA_OPTS
+RUN . /tmp/set_java_proxy.sh && echo $JAVA_OPTS
+
+RUN mv /usr/local/bin/run-jnlp-client /usr/local/bin/openshift-run-jnlp-client
+COPY ods-run-jnlp-client.sh /usr/local/bin/run-jnlp-client
```
faust2199, line 78: I believe we still have to set the script to executable after copying it to its destination.

michaelsauter: Let me double check, but I think if the source files have the executable bit set, `COPY` keeps this as-is.

jenkins/agent-base/Dockerfile.rhel8
```diff
@@ -0,0 +1,87 @@
...
+    && snyk --version \
+    && echo 'Snyk installation completed!'; \
+    fi
+
+# Set java proxy var.
+COPY set_java_proxy.sh /tmp/set_java_proxy.sh
+RUN . /tmp/set_java_proxy.sh && echo $JAVA_OPTS
+
+# Customize entrypoint.
+RUN mv /usr/local/bin/run-jnlp-client /usr/local/bin/openshift-run-jnlp-client
```
faust2199, line 70: Here the original executable is renamed to "openshift-run-jnlp-client", however in ods-run-jnlp-client.sh it is referenced by "ocp-run-jnlp-client". We need to use the same name in both places.

jenkins/agent-base/Dockerfile.rhel7
faust2199: Verified that the permission was right. No need to chmod here.

jenkins/agent-base/Dockerfile.rhel8
michaelsauter: :( Fixed in 9f121cc5ac651f220c27f459f82a9533c7af050a.

michaelsauter:
I'm opening this for review now, as Go, Node, Python, Java work on top of this.

I'm still not sure / happy with the naming of the `Dockerfile`. Should it be `Dockerfile.rhel8` or should we use `Dockerfile.ubi8` instead? `Dockerfile.rhel8` is used by https://github.com/openshift/jenkins/tree/master/slave-base, but otherwise it feels more natural to me to use `Dockerfile.ubi8` because it is meant to be used on top of the UBI image.

Preferences?

clemensutschig:
@michaelsauter  - `ubi8`

clemensutschig (approved):

PR #885, opened 2020-11-17
Feature/add cluster admin groups as project admin on project provisioning

stitakis:


felipecruz91:
@stitakis Just to double-check, do you think it is needed to pass the env var as parameter to the `/create-projects/create-cd-jenkins.sh` script?

jenkins/ocp-config/deploy/jenkins-master.yml
```diff
@@ -198,6 +200,8 @@ objects:
               value: '${DOCKER_REGISTRY}'
             - name: BITBUCKET_URL
               value: '${BITBUCKET_URL}'
+            - name: CLUSTER_GROUPS
```
felipecruz91, line 203: @stitakis I'd call it `ODS_CLUSTER_GROUPS` to be consistent with line 65

stitakis: sure, np

stitakis:
@felipecruz91 not sure what you mean with _"do you think it could also be required to pass the env var as parameter to the /create-projects/create-cd-jenkins.sh script?"_

jenkins/ocp-config/deploy/jenkins-master.yml
```diff
@@ -62,6 +62,8 @@ parameters:
   required: true
 - name: REPO_BASE
   required: true
+- name: ODS_CLUSTER_GROUPS
```
michaelsauter, line 65: If I understand it correctly, then the cluster groups are meant for provisioning. If this is correct, then the env variable should not be added here, but be passed as a parameter to the pipeline from the provisioning app.

configuration-sample/ods-core.env.sample
```diff
@@ -189,6 +189,9 @@ JENKINS_AGENT_BASE_SNYK_DISTRIBUTION_URL=https://github.com/snyk/snyk/releases/d
 # You may also point to repository underneath REPO_BASE.
 SHARED_LIBRARY_REPOSITORY=https://github.com/opendevstack/ods-jenkins-shared-library.git
 
+# Comma-separated list of group type and group name to be added as role to a project on project provision (see create-project.sh)
+#ODS_CLUSTER_GROUPS=ADMINGROUP=<YOUR_CLUSTER_ADMIN_GROUP>
```
michaelsauter, line 193: Shouldn't be commented - let's have the example in a comment, and the default be the empty value.

michaelsauter (changes requested):

jenkins/ocp-config/deploy/jenkins-master.yml
felipecruz91: @michaelsauter Could you point in the code where I should make that change?

stitakis: @michaelsauter we are bringing a new aspect of the provisioning process. This change provides a way to setup project permission to admin groups independently whether or not the create project pipeline was triggered by provisioning app. I'm highlighting this just to be sure that we are on the same page. If this is a good idea or not is something that we can discuss. What do you think?

michaelsauter: OpenShift Project provisioning is done through https://github.com/opendevstack/ods-core/blob/master/create-projects/Jenkinsfile. If we add a new parameter to this pipeline then the thing which triggers the pipeline (right now the prov app) should send the parameter. I don't want to have this environment variable as part of every Jenkins deployment.

stitakis: fair enough, thanks for the feedback! @felipecruz91 we need to move the parameter to the ProvApp!

felipecruz91: Understood, thank you @michaelsauter. I will coordinate with @stitakis to move it to the prov app

stitakis:
as we following a different approach this change is not needed anymore. I'll close it

PR #892, opened 2020-11-24
Install glibc-langpack-en

michaelsauter:
Avoids "warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)"

See https://github.com/AdoptOpenJDK/openjdk-docker/pull/265.

### DIFF
--- a/jenkins/agent-base/Dockerfile.ubi8
+++ b/jenkins/agent-base/Dockerfile.ubi8
@@ -11,7 +11,10 @@ ENV SONAR_SCANNER_VERSION=3.1.0.1141 \
 ARG APP_DNS
 ARG SNYK_DISTRIBUTION_URL
 
-RUN yum -y install openssl \
+# Add CentOS 8 repositories.
+COPY yum.repos.d/centos8.repo /etc/yum.repos.d/centos8.repo
+
+RUN yum -y install glibc-langpack-en openssl \
     && yum clean all \
     && rm -rf /var/cache/yum/*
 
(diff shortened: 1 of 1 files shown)


michaelsauter:
@oalyman @gerardcl @henrjk @faust2199 Fixed the build failure by adding the CentOS 8 repo right in the beginning. This is not needed when pulling from registry.redhat.io and building inside an OpenShift cluster. I did not find out why the quay.io based image cannot find all those repos, I was under the impression that the UBI repos are minimal, but available to everyone. Maybe someone else wants to take a look and see what I am missing? In the meantime though, I think with this PR we have a fully working agent image so from my side we should go ahead with this for now.

faust2199 (approved):
LGTM. It appears images from quay.io only have rhel-* repos enabled. If we don't add other repos (ubi8 or centos8) manually, they will only work in environments with RHEL subscription.

michaelsauter:
@faust2199 Nice find. I checked in one image, and you are right. I don't really get why this is the case - if anything, I'd expect the quay.io image to work better without a subscription. I tried to circumvent this by copying `ubi.repo` to my local filesystem, then:

```
COPY yum.repos.d/ubi.repo /tmp/ubi.repo
RUN if [ -f /etc/yum.repos.d/ubi.repo ]; then echo 'found ubi.repo'; else echo 'move ubi.repo' \
    && mv /tmp/ubi.repo /etc/yum.repos.d/ubi.repo \
    && yum-config-manager --setopt=\*.skip_if_unavailable=1 --save; \
    fi
```

But this still fails. The error is now:
```
Error: Invalid version flag: if
The command '/bin/bash -o pipefail -c yum -y install glibc-langpack-en openssl     && yum clean all     && rm -rf /var/cache/yum/*' returned a non-zero code 1
```

So I guess we leave the version proposed by this PR, and come back to this later to better understand what is going on.

PR #896, opened 2020-11-25
adds configuration sample, extends prov app cm.yml with default project groups

stitakis:
completes https://github.com/opendevstack/ods-provisioning-app/issues/635 (port to master)


stitakis:
One check is failing because quay.com return a 5XX error, we need to wait until the server is back again. This is the error:
```
Run ./.github/workflows/build-docker-image.sh \
Sending build context to Docker daemon  16.38kB

Step 1/30 : FROM quay.io/openshift/origin-jenkins-agent-base
Get https://quay.io/v2/openshift/origin-jenkins-agent-base/manifests/latest: received unexpected HTTP status: 502 Bad Gateway
Error: Process completed with exit code 1.
```

michaelsauter (approved):
We can still go ahead and merge from my point of view as the failure is definitely unrelated.

PR #903, opened 2020-11-27
Install Helm

### LINKED ISSUE #551: Add helm/kubectl to tools
The current way ODS builds and deploys to OpenShift is quite opinionated. That makes it hard to deploy more demanding infrastructure.

I would like to be able to use custom deployment descriptors that are under my control when deploying to OpenShift.

OpenShift 4 is converging towards more "standardized" kubernetes resources (e.g. planning to remove DeploymentConfig in favor of Deployment in the future). This makes "native" (as in kubernetes native) tools a prime target to be available.

Helm is one of the tools where a large community exists and already provides lots of advice and documentation. Since the advent of helm v3 there's no need for privileged components (like tiller) in the cluster. 

**Describe alternatives you've considered**
tailor and OpenShift templates. Both seem insufficient when dealing with more complex environments that should be able to get deployed from a single source of truth.

**Additional context**

michaelsauter:
Also installs two helpful plugins, helm-diff and helm-secrets. Other plugins might be added depending on user need.

Closes #551.

This also installs `helm` into the RHEL7 image, even though there is no full support for OpenShift 3.11. That said, it might still be helpful for some use cases (e.g. deploying elsewhere), and it keeps the two images in sync. I'm not planning to backport this change to 3.x.

How OpenDevStack will make use of Helm directly has not been decided yet, but it is likely to be integrated, see https://github.com/opendevstack/ods-jenkins-shared-library/issues/495.

FYI @serverhorror @segfault16

### DIFF
--- a/jenkins/agent-base/Dockerfile.ubi8
+++ b/jenkins/agent-base/Dockerfile.ubi8
@@ -5,6 +5,9 @@ SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 ENV SONAR_SCANNER_VERSION=3.1.0.1141 \
     CNES_REPORT_VERSION=3.2.2 \
     TAILOR_VERSION=1.3.0 \
+    HELM_VERSION=3.4.1 \
+    HELM_PLUGIN_DIFF_VERSION=3.1.3 \
+    HELM_PLUGIN_SECRETS_VERSION=3.3.0 \
     GIT_LFS_VERSION=2.6.1 \
     JAVA_HOME=/usr/lib/jvm/jre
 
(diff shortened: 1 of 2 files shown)


jenkins/agent-base/Dockerfile.ubi8
```diff
@@ -39,10 +40,22 @@ RUN cd /tmp \
...
+# Install Helm.
+RUN cd /tmp \
+    && mkdir -p /tmp/helm \
+    && curl -LO https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
+    && tar -zxvf helm-v${HELM_VERSION}-linux-amd64.tar.gz -C /tmp/helm \
+    && mv /tmp/helm/linux-amd64/helm /usr/local/bin/helm \
+    && chmod a+x /usr/local/bin/helm \
+    && helm version \
+    && helm plugin install https://github.com/databus23/helm-diff --version v3.1.3 \
+    && helm plugin install https://github.com/jkroepke/helm-secrets --version v3.3.0 \
```
gerardcl, line 57: should we make these plugins versions also ENVs?

oalyman (dismissed):

gerardcl (dismissed):
aside of my comment, LGTM! 🚀

jenkins/agent-base/Dockerfile.ubi8
michaelsauter: Why not, added in 5d93086.

gerardcl (approved):
🚀

PR #904, opened 2020-11-27
Refactor test targets

### LINKED ISSUE #792: Make test dependencies explicit, and remove duplicated tests
The recent refactoring of the quickstarter tests, and issues like #790 got me thinking about our tests. First, here's what we have right now as I understand it:

1. "Verify tests". These check that an actual ODS installation is as it should be.
2. "Create project tests". These (re)create `unitt-*` projects, in 3 different ways: once just the projects through shell scripts, once projects with Jenkins through shell scripts, and once projects with Jenkins through the webhook proxy. The last test actually depends on the verify tests as the webhook proxy must exist. Originally, this test package was designed to be run in a naked, local OpenShift cluster (@hugowschneider correct me if I'm wrong?).
3. "Quickstarter tests". These are executed in the `unitt-*` projects (from point 2 above - the project does not cleaned up after the tests so still exists for the quickstarter tests), and execute quickstarters through the webhook proxy in the `unitt-cd` project. They also depend on point 1 above as they require e.g. a working SonarQube.

A couple of things that I observed:

* The tests in (2) are to a large extend redundant. There is just one additional check on top of what (1) already provides, which could easily be moved to (1): `CheckProjectSetup` in `create-projects_test.go`. The only difference I see is that the first two tests in (2) in theory work without an actual ODS installation. I think I would propose to only run one test, which works without ODS installation, and kill the other two.
* Right now there is an implicit dependency of the quickstarter tests on both (1) and (2) - it is only noted in the readme. I think it would be good to have some kind of explicit prerequisite, e.g. create `unitt-*` projects if they don't exist, and check if they are configured correctly ... here we can just use the code already there in (1).
* The only thing in (1) which is not required for the quickstarters is the provisioning app check (the odsverify project). Maybe it would be good to not execute the prov app check if we only want to run quickstarter tests to save time.

So all up, my current thinking is:

* Kill webhook proxy and project-only tests from `create-projects`
* Move shell-script approach to (re)create `unitt-*` projects and check config of it to a prerequisite of the quickstarter test
* Move prov app verification into it's own target

With this in mind, the new targets would look sth. like this:
```
verify: verify-ods-namespace verify-provisioning-app

verify-ods-namespace:
  # what we have currently in verify, minus prov app

verify-provisioning-app:
  # prov app

prepare-quickstarter-test:
  # unitt project creation if it does not exist, plus config check (one test from the current create-projects tests)

test-quickstarter: verify-ods-namespace prepare-quickstarter-test
  # quickstarter tests as is right now
```

I'm not entirely sure yet if this is the right thing to do, but I wanted to write down my thoughts.

@henrjk @hugowschneider @clemensutschig @georgfedermann @segator What are your thoughts?

michaelsauter:
Closes #792.

This PR removes the project/component provisioning from "make verify",
and makes it available under a new target "make smoketest" to signal that this actually creates something. "verify" does not make any change now - it just checks if all is configured as it should be.

Further changes:

* Fixes issue with sourcing envs in Nexus/SonarQube test (could be an issue with bash 3.2?)
* Add "make help" target
* Add "test-create-projects" target to just run those tests which are required to setup the `UNITT` project


michaelsauter:
Just realised I forgot to update the `README.md`. Will update this as well.

henrjk (commented):
Here is an experiment from my mac os system:

In the parent folder of ods-configuration, I have script `nexus-env-shebang.sh`:

```bash
#!/usr/bin/env bash
set -x
echo $BASH_VERSION
(echo $BASH_VERSION)
source <( grep NEXUS ods-configuration/ods-core.env.sample) 
echo $NEXUS_HOST
```

and `nexus-bin-shebang.sh`:

```bash
#!/bin/bash
set -x
echo $BASH_VERSION
source <( grep NEXUS ods-configuration/ods-core.env.sample) 
echo NEXUS_HOST
```

When I run them I get the following output:

```
√ ods $ ./nexus-bin-shebang.sh 
+ echo '3.2.57(1)-release'
3.2.57(1)-release
+ source /dev/fd/63
+ echo NEXUS_HOST
NEXUS_HOST
++ grep NEXUS ods-configuration/ods-core.env.sample
√ ods $ ./nexus-env-shebang.sh 
+ echo '5.0.11(1)-release'
5.0.11(1)-release
+ echo '5.0.11(1)-release'
5.0.11(1)-release
+ source /dev/fd/63
++ grep NEXUS ods-configuration/ods-core.env.sample
++ NEXUS_HOST=nexus-cd.192.168.56.101.nip.io
++ NEXUS_URL=https://nexus-cd.192.168.56.101.nip.io
++ NEXUS_USERNAME=developer
++ NEXUS_PASSWORD=changeme
++ NEXUS_PASSWORD_B64=changeme
++ NEXUS_AUTH=developer:changeme
+ echo nexus-cd.192.168.56.101.nip.io
nexus-cd.192.168.56.101.nip.io
```

So the previous version of this code worked on my Mac with the bash shell I installed with brew. However it does not work with the standard mac bash. 

It is not clear to me what shell version we should expect. However it appears that targeting some more recent bash version would makes sense. @michaelsauter What do you think?

michaelsauter:
@henrjk Thanks for following this up. That explains things!

Personally, I would like to stay compatible with the old bash version:

* We are dealing with tests here, and I would like to have it as easy as possible to execute them. Having to potentially update your bash just to run tests is too big of a hurdle in my eyes. This view is also driven by personal experience: I spend a lot of time in the terminal, but I use zsh. Therefore I don't really care about my bash version, which means I'm using the version that ships with macOS, which is ancient. I would wait with moving to a newer bash version until macOS ships with a newer one by default.
* The proposed change in this PR is really small so it does not hurt much to make the script compatible with older bash versions
* The error given if you do not have a recent enough bash is quite cryptic - it complains about unset variables, which would take someone a lot of time until they figure out it is because bash is too old.

clemensutschig (approved):

henrjk:
@michaelsauter I see your point.

However I don't think Apple will ever upgrade bash. If anything they will remove it, which means that everybody would have to install their own bash. I would guess this will not happen anytime soon.

I would prefer if we can assume a newer bash version than the ancient MacOS version. True users who use zsh would have to install a bash that you'd not really use. On the other hand there can be incompatibilities between zsh and bash I suppose and it might be useful to have bash to ensure compatibility.

In case you'd want to make the changes perhaps the shebang should be changed to `/bin/bash`

michaelsauter:
@henrjk Sorry, didn't see your comment before merging. I hope you are OK with the changed way to read the env files ...

PR #922, opened 2020-12-14
fixes freeze of systemctl enable docker

stitakis:
Installation of odsbox is failing because the activation of docker in systemctl freezes.

This problem seems to happen since a new version of the docker package is available.

To fix this problem I'm fixing the docker version to install to the latest one that worked.

I tested running packer locally for the branch bugfix/ods-devenv-build-fixes and also master.


clemensutschig (approved):

stitakis:
@clemensutschig packer AMI build of master branch was successful on my notebook... I'll port back this change to branch 3.x

PR #932, opened 2021-01-04
Reference oc Git commit 3.9 instead of yanked tag

### LINKED ISSUE #931: AMI build fails during "make prep-tools"
```
ESC[0;32m    amazon-ebs: which go-junit-report || go get github.com/jstemmer/go-junit-reportESC[0m
ESC[1;31m==> amazon-ebs: which: no go-junit-report in (/usr/local/bin:/usr/bin:/home/openshift/bin:/home/openshift/go/bin)ESC[0m
ESC[1;31m==> amazon-ebs: go: github.com/openshift/api@v3.9.0+incompatible: reading github.com/openshift/api/go.mod at revision v3.9.0: unknown revision v3.9.0ESC[0m
ESC[1;31m==> amazon-ebs: make: *** [prep-tools] Error 1ESC[0m
2021/01/04 01:02:22 packer-builder-amazon-ebs plugin: [ERROR] Remote command exited with '2': /tmp/script_9696.sh  --branch master
2021/01/04 01:02:22 packer-builder-amazon-ebs plugin: [INFO] RPC endpoint: Communicator ended with: 2
```

I'm going to look into this ... 

FYI @clemensutschig @georgfedermann

michaelsauter:
`github.com/openshift/api` has yanked the `v3.9.0` tag from the repo, which
causes Go to fail when trying to download the dependency. I'm guessing they did
this because 3.9 was pre-modules. Anyway, I ran:

```
go get github.com/openshift/api@release-3.9
```

This records the last commit of the `release-3.9` branch
(https://github.com/openshift/api/commits/release-3.9) in `go.mod`,
which should work again (looking at the hash in `go.sum`, it appears to point to a different commit then before though ...).

Fixes #931.

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/jstemmer/go-junit-report v0.9.1
 	github.com/kr/pretty v0.2.1 // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/openshift/api v3.9.0+incompatible
+	github.com/openshift/api v0.0.0-20180801171038-322a19404e37
 	github.com/openshift/client-go v3.9.0+incompatible
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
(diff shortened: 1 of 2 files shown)


clemensutschig (commented):
Is there no 3.11?

michaelsauter:
@clemensutschig Ah right, did not think of it. There is a `release-3.11` branch, yes. However, we have other, related dependencies that have been locked ... I think for now I'd just like to unblock the tests, and then do the bump separately once we have a green build (there could be other failures further down right now that we do not know about yet).

clemensutschig (approved):

Issue #936, opened 2021-01-05
ODS in a Box


felipecruz91:
The concept of ODS in a box it's a really great idea and I am learning more about it in my spare time. I am interested in creating a cloud-agnostic VM image (not coupled to AWS for instance) that can run on my laptop (not in the cloud) using [Vagrant](https://www.vagrantup.com/) with VirtualBox.

I've forked `ods-core` (master branch) into my personal account, made the pertinent changes, and run `packer build -on-error=ask ./ods-devenv/packer/CentOS2ODSBoxVagrant.json`

It installs the Atlassian stack, OpenShift, etc but it fails when it tries to build the nexus image:

**Screenshots**

![nexus-build-error](https://user-images.githubusercontent.com/15997951/103633129-99680180-4f45-11eb-9cc6-78a9971085a0.PNG)

![nexus-build-events](https://user-images.githubusercontent.com/15997951/103633310-da601600-4f45-11eb-8514-de9dec66aa02.PNG)

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack (master branch)

**Additional context**
From the error message, I understand that the user does not have the RBAC permissions needed to list the secrets under the ods namespace.

felipecruz91:
After this failure, I have triggered the nexus build with `oc start-build nexus -n ods` and now it has completed successfully. I _guess_ it is a timing issue where the secrets are not created by the time the build starts.

Perhaps it would be interesting to wait for all the resources to be created in OpenShift before starting the [setup_nexus](https://github.com/felipecruz91/ods-core/blob/feature/ods-box/ods-devenv/scripts/deploy.sh#L1887) function.

clemensutschig:
Do you have the err message from the failed run?

Felipe Cruz Martinez <notifications@github.com> schrieb am Di., 5. Jän.
2021, 11:48:

clemensutschig:
Can you try this again with a wait/sleep before the startbuild?

Felipe Cruz Martinez <notifications@github.com> schrieb am Di., 5. Jän.
2021, 11:22:

felipecruz91:
@clemensutschig It seems to be a timing issue, yes.

In this second run, it has progressed a lot and has started testing the ODS installation but it seems that the Prov App is not running:

```cli
    vagrant: === RUN   TestVerifyOdsProjectProvisionThruProvisionApi
[...]

PR #939, opened 2021-01-07
Update start-and-follow-build.sh

felipecruz91:
Refers to #936

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Add SonarQube readiness probe ([#495](https://github.com/opendevstack/ods-core/pull/495))
 
 ### Changed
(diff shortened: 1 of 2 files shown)


clemensutschig (changes requested):
Changelog missing. And more importantly... also failed needs to be Checked for..

michaelsauter:
I think the actual problem is slightly elsewhere. In the issue, I linked to the log output. The `start-build` command "errors" because it has a built-in timeout in which the build needs to progress from "pending" to "running", which is exceeded in your case. It seems that when this timeout occurs, the command still exits with 0 and the next checks see the build as pending. 

Your change hides the build log output when the `start-build` timeout is exceeded ... maybe better to 1) start 2) wait 3) display logs? Though I'd be sad to have no streaming output just for the sake of catering for slow systems.

clemensutschig (approved):

PR #950, opened 2021-02-10
Fix PVC creation on OpenShift 4

michaelsauter:
Otherwise the PVC resources are stuck in "pending".

The old way worked on an Azure RedHat Openshift 4.5, but fails on RedHat Dedicated OpenShift 4.6.

Note that the SonarQube PVC resources are already defined with `storageClassName`, and without the annotation.

Will open a PR for `master` as well.

### DIFF
--- a/nexus/ocp-config/Tailorfile
+++ b/nexus/ocp-config/Tailorfile
@@ -4,5 +4,6 @@ param-file ../../../ods-configuration/ods-core.env
 ignore-unknown-parameters true
 preserve-immutable-fields true
 preserve pvc:/spec/volumeMode
+preserve pvc:/metadata/annotations/volume.beta.kubernetes.io/storage-class
 
 dc,is,pvc,route,svc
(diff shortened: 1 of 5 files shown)


gerardcl (approved):
LGTM!
question: is then `STORAGE_CLASS_DATA` and `STORAGE_CLASS_BACKUP` required from ods-configuration PoV?

michaelsauter:
@gerardcl those two params are part of ods-configuration already, and are just used in a different resource field here. No change to the values itself.

gerardcl:
I mentioned that because usually from a cluster point of view it is expected to not define the classes in order to ensure using the cluster default one. Not sure if that applies to any openshift cluster type though.

michaelsauter:
Yea, we discussed this several times already :) I'm not convinced that you'd always want to use the default. I'd rather allow users to control this (at the cost of having to control this - but then again this is a one time thing to do so not a big issue).

clemensutschig:
@michaelsauter way fwd?

michaelsauter:
@clemensutschig Thanks for the reminder. I'll close and re-open for master. We need this change for v4.0.0.

Issue #966, opened 2021-03-25
Sonar edition not set in Jenkins master DeploymentConfig


felipecruz91:
**Describe the bug**
SonarQube won't analyze non-master branches unless the env. var `SONAR_EDITION` is set to `enterprise` in the  Jenkins master **DeploymentConfig** resource in OpenShift.

**Expected behavior**
Despite defining the `SONAR_EDITION=enterprise` line in the ods-core.env file, it is not set here: https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/deploy/jenkins-master.yml#L53-L56, which means the env. var is not being used in this case.

**Screenshots**

![image](https://user-images.githubusercontent.com/15997951/112460651-6c1a3b00-8d5f-11eb-80b6-c657f7b32b9b.png)

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack 3.1

michaelsauter:
I'm closing this as per #967.

PR #967, opened 2021-03-25
Added SONAR_EDITION env. var.

felipecruz91:
Refers to #966

### DIFF
--- a/jenkins/ocp-config/deploy/jenkins-master.yml
+++ b/jenkins/ocp-config/deploy/jenkins-master.yml
@@ -196,7 +196,7 @@ objects:
                 secretKeyRef:
                   key: auth-token
                   name: sonarqube
-            - name: SONAR_EDITION
+            - name: SONARQUBE_EDITION
               value: '${SONAR_EDITION}'                  
             - name: SHARED_LIBRARY_REPOSITORY
               value: '${SHARED_LIBRARY_REPOSITORY}'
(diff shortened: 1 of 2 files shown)


jenkins/ocp-config/deploy/jenkins-master.yml
```diff
@@ -193,6 +196,8 @@ objects:
                 secretKeyRef:
                   key: auth-token
                   name: sonarqube
+            - name: SONARQUBE_EDITION
```
felipecruz91, line 199: @renedupont @michaelsauter Note in `ods-core.env` the name of the variable is `SONAR_EDITION` but *within* Jenkins it is `SONARQUBE_EDITION`.

michaelsauter:
This should not be needed. The env var is set in the image: https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/build/bc.yml#L110.

It was done that way as in the first version of ODS, the env var was not part of either DC or image. To add it in a newer version without requiring users to change their DC, it was added in the image.

braisvq1996:
@michaelsauter If I check the current SonarQube edition in the Jenkins pipeline, this is what we get:

image

image

braisvq1996:
It may be because the env value in Jenkins needs to be SONARQUBE_EDITION instead of SONAR_EDITION ?

image

Env values in Jenkins master:

image

michaelsauter:
@braisvq1996 good point. Then this is a bug and we need to align the variable naming. Can you update this PR accordingly?

felipecruz91:
@michaelsauter As @braisvq1996 has identified a workaround that needs to be done in the shared lib, I'll be closing this PR and opening a new one in the shared lib repository.

Issue #975, opened 2021-04-21
APEX code analyzer plugin for SonarQube


braisvq1996:
**Is your feature request related to a problem? Please describe.**
We would like for SonarQube to be able to scan Apex code

**Describe the solution you'd like**
Install APEX code analyzer plugin into SonarQube

**Describe alternatives you've considered**
No alternatives considered

**Additional context**
We can submit a PR regarding this for master and 3.x branches

@metmajer @felipecruz91 @segator @borja44

braisvq1996:
This plugin is a requirement for the Salesforce quickstarter

clemensutschig:
this is waiting for 3.x correct?

braisvq1996:
I am currently preparing the 3.x PR, in a moment I will create it

PR #977, opened 2021-04-22
Add sonar-apex-plugin for SonarQube Enterprise and Datacenter editions

braisvq1996:
This is the initial PR for the issue #975

There is an issue where this plugin gives errors when adding it to Sonar Community and Developer editions.
This plugin can only be added to Enterprise and Datacenter editions
https://docs.sonarqube.org/8.2/analysis/languages/overview/

That is why we had to make a difference between the editions and add an if statement into the dockerfile

If you have any thoughts on this please share them

@metmajer @felipecruz91 @michaelsauter @gerardcl

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -71,6 +71,7 @@ jobs:
     strategy:
       matrix:
         version: ['7.9', '8.2.0.32929'] # 7.9 = LTS, 8.2 = latest version
+        edition: ['community', 'enterprise']
     steps:
       -
         name: Checkout repository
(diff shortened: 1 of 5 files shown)


clemensutschig (changes requested):
there is a test missing ... where plugins are checked

felipecruz91:
@braisvq1996 @clemensutschig The SonarQube tests are failing as the test suite uses a [non-commercial distribution](https://github.com/opendevstack/ods-core/blob/30f3966cc80df9078a2a853d58581da024f97539/sonarqube/test.sh#L56) of SonarQube.

Therefore, the `apex:1.8.2.1946` plugin shall be removed from the `expectedPlugins` [list](https://github.com/opendevstack/ods-core/blob/30f3966cc80df9078a2a853d58581da024f97539/sonarqube/test.sh#L193).

michaelsauter:
@felipecruz91 Would be great if we can have a commercial edition under test as well. I suggest to try what you suggested in a separate PR, and then have this PR be based on that.

clemensutschig:
@braisvq1996  - do you fix this one now as well

braisvq1996:
@clemensutschig - Yeah, that is the plan.

clemensutschig (approved):

PR #989, opened 2021-06-09
Feature/aqua configmaps

jorge-romero:
This PR contains two topics:
### Create needed configmaps in the ods namespace to control the behavior of the SharedLibrary with aqua, and the neede configuration for the SL to be able to scan it

1. Aqua enablement at cluster level
2. Aqua url
3. Aqua internal registry that will be used to store the information of the scaned images
4. Secret name to access aqua instance
5. Comma separated list of email addresses to provide feedback in case there are problems using Aqua
6. The name of a Nexus repository to store the reports generated by Aqua

### The script to install aqua and the instructions install them in a odsbox.
The script and all the needed resources to install aqua in the box can be found under ods-devenv/scripts/aqua.
By default Aqua won't be installed in the odsbox, but the scripts are provided to install it and allow anyone to develop new features related with Aqua in the box.


metmajer:
@jorge-romero let us know when you are ready for a review

Makefile
```diff
@@ -42,6 +42,10 @@ prepare-config:
 install-ods-project:
 	cd ods-setup && ./setup-ods-project.sh --namespace $(ODS_NAMESPACE) --reveal-secrets
 
+# AQUA SETUP
+## Setup central "ods" project.
+setup-aqua-configmap: install-ods-project
```
michaelsauter, line 47: I'd probably not use the dependency on `install-ods-project`.

jorge-romero: it was a suggestion for another team member but I do agree with you, those are topics that can be executed independently, I'll remove it. Thanks for your input

jorge-romero:
FYI. @metmajer , @michaelsauter 

There is currently a bug with the alpine image used to build the sonarqube image:

https://github.com/AdoptOpenJDK/openjdk-docker/issues/353

Currently is not possible to pass the tests.

Perhaps a good option to prevent this kind of failures will be to set a specific image for each of the components of ODS, that had to be upgraded periodically to avoid vulnerabilities problems, but won't break the process of developing new feature in ODS.

metmajer:
@jorge-romero I don't understand. Can you provide more insights on the problem you're facing? What's happening?

michaelsauter:
Previous successful builds are based on https://hub.docker.com/layers/adoptopenjdk/openjdk11/x86_64-alpine-jre-11.0.10_9/images/sha256-a97add2e5cc7588fb7bf67a9be4815dc0acabe914af123fbac4e028a0c3140bf?context=explore. The failing builds pulled https://hub.docker.com/layers/adoptopenjdk/openjdk11/x86_64-alpine-jre-11.0.11_9/images/sha256-a5f72791e32e1d3493f602bde395508147e35e42669bd4a014cc3b87a9f918a6?context=explore.

The build failure is completely unrelated to this PR.

@jorge-romero's solution in https://github.com/opendevstack/ods-core/pull/989/commits/8e8e4a22f0f0ecad0f28cb8c1dfdac134ee0afc2 pins the tag to the working one. I think this is a good solution for now but we should monitor this so that we don't fall behind new versions.

michaelsauter:
@jorge-romero it would be nice to extract that fix into a separate PR and provide the context so that it is listed separately from this PR in the changelog.

jorge-romero:
> @jorge-romero I don't understand. Can you provide more insights on the problem you're facing? What's happening?

The existing problem is that when the SonarQube image is being built, it always fails. 
SonarQube image is based on AdoptJDK11, that it is based on the Alpine Docker distribution. For the latest one there has been several changes that are incompatible with the installation of SonarQube, there are some glibc conflicts that fails the build execution while it is installing the needed packages. 
That was reported already in the AdoptJDK github issue tracker AdoptOpenJDK/openjdk-docker#353 and it is stated that it is not a problem of AdoptJDK, is a problem of the Alpine distribution.

As Michael said, the fast solution is to set a previous version of the AdoptJDK image that we are using, otherwise it will fail until it is fixed by Alpine/AdoptJDK projects.

I had faced this kind of problems before and we follow the following rules:
- Always set a specific version of the base images
- Periodically increase the version to avoid security problems and get enhancements in the images.

We did it in that way as we don't control what's in the source images of the docker containers (FROM xyz..) so for automatic builds we avoided in this way having a system that is not able to recreate the images.

jorge-romero:
> @jorge-romero it would be nice to extract that fix into a separate PR and provide the context so that it is listed separately from this PR in the changelog.

Ok I'll move it to another branch I'll assign you both as reviewers to get that in place as soon as possible so it won't break the pipeline for any other features developed

jorge-romero:
@michaelsauter, as you mention, I have remove it from here the fix of sonarqube, you could find it on https://github.com/opendevstack/ods-core/pull/994

jorge-romero:
It worked, I was able to build a box with the changes introduced in the repositories, thank you. This pr can be merged.

configuration-sample/ods-core.env.sample
```diff
@@ -369,3 +369,27 @@ ADDITIONAL_JIRA_PROJECT_TEMPLATE_PROPERTIES="# ---------------------------------
...
 # can be overwritten by specifying the param in this file.
+
+########
+# Aqua #
+########
+# Is enabled of disabled aqua at cluster level
+AQUA_ENABLED=false
+
+# Is enabled for projects Aqua
+AQUA_ENABLED_FOR_PROJECT=true
```
metmajer: @jorge-romero @hrcornejo in our new implementation, a project is disabled selectively in the cluster-level ConfigMap. Is this configuration parameter still valid?

jorge-romero: That's true, it is not valid anymore, I have already remove it as the project level configuration done in ODS namespace will be done by the cluster admins.

metmajer:
Then, let's ask @michaelsauter for an approval.

jorge-romero:
@michaelsauter, Could you have a look at this PR and approve it if you consider that it is right?

Makefile
```diff
@@ -42,6 +42,10 @@ prepare-config:
 install-ods-project:
 	cd ods-setup && ./setup-ods-project.sh --namespace $(ODS_NAMESPACE) --reveal-secrets
 
+# AQUA SETUP
+## Setup central "ods" project.
```
michaelsauter, line 46: This text is displayed when you run `make help`. I think it should mention Aqua here.

configuration-sample/ods-core.env.sample
```diff
@@ -369,3 +369,24 @@ ADDITIONAL_JIRA_PROJECT_TEMPLATE_PROPERTIES="# ---------------------------------
...
+########
+# Aqua #
+########
+# Enable or disable Aqua at the cluster level
+AQUA_ENABLED=false
+
+# The URL to the Aqua service
+AQUA_URL=
+
+# The name of a container image registry in Aqua
```
michaelsauter, line 382: Maybe give an example here? Is this host only? or a URL with scheme?

configuration-sample/ods-core.env.sample
```diff
@@ -369,3 +369,24 @@ ADDITIONAL_JIRA_PROJECT_TEMPLATE_PROPERTIES="# ---------------------------------
...
+AQUA_REGISTRY=
+
+# The name of a secret containing credentials (username and password) of an Aqua user that has scanner rights
+AQUA_SECRET_NAME=
+
+# List of email addresses to provide feedback in case there are problems using Aqua
```
michaelsauter, line 388: what's the format? Comma separated?

ods-devenv/scripts/aqua/install-aqua.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
michaelsauter, line 1: I'd recommend at least `set -eu` on any script so that it fails on error. Enabling pipefail could also be considered.

configuration-sample/ods-core.env.sample
```diff
@@ -369,3 +369,24 @@ ADDITIONAL_JIRA_PROJECT_TEMPLATE_PROPERTIES="# ---------------------------------
...
+AQUA_ALERT_EMAILS=
+
+# The name of a Nexus repository to store the reports generated by Aqua
+AQUA_NEXUS_REPOSITORY=
```
michaelsauter, line 392: Should we default this to `leva-documentation`?

michaelsauter (approved):
Looks good to me.

I can't really assess the `ods-devenv` changes as I don't know the details of the box and haven't tried setting up Aqua inside.

Makefile
jorge-romero: You can find that we just added a better description for this goal in the latest commit

configuration-sample/ods-core.env.sample
jorge-romero: Added an example as we have in the odsbox after installing there Aqua

configuration-sample/ods-core.env.sample
jorge-romero: Change comment to specify that it is a comma separated list

configuration-sample/ods-core.env.sample
jorge-romero: Added as default leva-documentation unless another one is specified

ods-devenv/scripts/aqua/install-aqua.sh
jorge-romero: This task is currently used manually if you need to install aqua in a box, if that is required I'll add that setting to the script.

configuration-sample/ods-core.env.sample
metmajer: @jorge-romero I strongly suggest you define any defaults inside the configuration. can you please change that? (simple, easy to understand, and no hidden assumptions)

jorge-romero: @metmajer This is the current configuration, unless there will be another one that should be set with a default value we will keep those in this way:

# Enable or disable Aqua at the cluster level
AQUA_ENABLED=false

# The URL to the Aqua service, f.e. using a local service in OCP http://<aqua-service>.aqua.svc.cluster.local:8080
# using a hosted instance of Aqua in OCP => aqua service route https://<aqua-route>.<public-cluster-url>/
AQUA_URL=

# # The name of a the internal registry defined in Aqua to keep the track of the images scanned
AQUA_REGISTRY=

# The name of a secret containing credentials (username and password) of an Aqua user that has scanner rights
AQUA_SECRET_NAME=aqua-user-with-password

# Comma separated list of email addresses to provide feedback in case there are problems using Aqua
AQUA_ALERT_EMAILS=

# The name of a Nexus repository to store the reports generated by Aqua
AQUA_NEXUS_REPOSITORY=leva-documentation

AQUA_URL depends on were Aqua is deployed and the url provided to access aqua
AQUA_REGISTRY is the internal registry configured in Aqua
AQUA_ALERT_EMAIL will depend on the recipients that will be decide to receive notifications

metmajer: @jorge-romero must have been me then. When I checked the file last time, no default values were in there, so I expected them to be hard-coded in the codebase. Thanks!

metmajer (approved):

jorge-romero:
@michaelsauter The code has been verified, and we have the approval from Martin, could you approve/merge this PR?

clemensutschig (approved):
I am approving this (with objection) ... we have to think about a smart way to NOT have ods-devenv provide stuff ods does NOT provide! @metmajer fyi

PR #994, opened 2021-07-01
Change SonarQube image base image to avoid build problems

jorge-romero:


metmajer (approved):
Thanks @jorge-romero and @michaelsauter!

PR #1036, opened 2021-09-08
Fix config check

michaelsauter:
Without `^`, the check detects missing params which are actually
comments. We only want to check lines that start with a parameter.

Will backport to 4.x


gerardcl (dismissed):
LGTM! 🚀 

maybe we could squash to just one commit?

Issue #1000, opened 2021-07-12
Jenkins master is not able to deploy agents


braisvq1996:
**Describe the bug**
After a pipeline is created and/or executed Jenkins is not able to successfully run the an ODS jenkins agent.

This is happening when using a Jenkins master build with the code from ods-core, master branch into Openshift 3.11

**To Reproduce**
1. Install ODS from master branch into Openshift 3.11
2. Create a new project
3. Add Quickstarters into the new project
4. See the issue in the pipelines

**Expected behavior**
To successfully run the ODS pipelines

**Screenshots**
![image](https://user-images.githubusercontent.com/26645694/125285512-04dac380-e31b-11eb-8b78-c9c6a7401280.png)

**Affected version (please complete the following information):**
 - OpenShift: 3.11
 - OpenDevStack master

**Additional context**
One thing we have found is that in the pod yml, the node selector is diferent from the one used in ODS 3.x and is the cause of this issue

**ODS 3.x**
```
nodeSelector:
    beta.kubernetes.io/os: linux
```

**ODS master**
```
nodeSelector:
    kubernetes.io/os: linux
```

felipecruz91:
OpenShift 3 is based on an older version of Kubernetes, which is not anymore directly supported since Kubernetes plugin version 1.26.0.

To get agents working for Openshift 3, add this Node Selector to your Pod Templates:

beta.kubernetes.io/os=linux


Source: https://plugins.jenkins.io/kubernetes/

braisvq1996:
If we make use of a lower version of the Kubernetes plugin there are other plugins that need to be downgraded as well

![image](https://user-images.githubusercontent.com/26645694/125408530-4ae76400-e3bb-11eb-9bbe-41234179f593.png)

braisvq1996:
For what I have seen there are 3 plugins needed to be changed/added for the pipelines to work:

**Kubernetes**  version _1.25.2_ (downgrade from _1.29.7_)
**Openshift Sync** version _1.0.45_ (downgrade from _1.0.46_)
**Pipeline Utility Steps** version _2.5.0_ (was installed in ODS 3 but not for ODS 4 )

clemensutschig:
@braisvq1996 install all them - and see if it works (what you can do - is to first install them via marketplace) and try this way ...

clemensutschig:
why don't we try with the ocp4 piece(s) on 3 ... maybe they have backwards compatibility?

braisvq1996:
[...]

PR #1023, opened 2021-08-04
Update plugins.rhel7.txt

felipecruz91:
Refers to https://github.com/opendevstack/ods-core/issues/1021

Updated plugins to match the current plugins versions present in the EDP Box.

PR #1038, opened 2021-09-09
add Dockerfile.rhel7 and Dockerfile.centos7

s2oBCN:
From https://github.com/opendevstack/ods-core/issues/1000

Problem:
- Thre's a compatibility problem with Dockerfile.rhel7 and Dockerfile.centos defined plugins.
- The initial name of Dockerfile.rhel7  used the Centos7 base, so it's renamed.

Solution:
Now we have a  Dockerfile.rhel7 and Dockerfile.centos7 with respectives plugins.rhel7.txt and plugins.centos7.txt definitions


michaelsauter:
Does this mean the issue you experience as described in #1000 happens in an OpenShift 3 cluster which is using CentOS images? Is this a test cluster? Why are you not using the RHEL7 base images? Or am I confusing things?

oalyman (approved):

s2oBCN:
> Does this mean the issue you experience as described in #1000 happens in an OpenShift 3 cluster which is using CentOS images? Is this a test cluster? Why are you not using the RHEL7 base images? Or am I confusing things?
> 
> Apart from that, the change itself looks fine for me 👍

It's in OC3 using RHEL7 base image.  The problem was  this change https://github.com/opendevstack/ods-core/pull/1023
The ODSBoxes uses the OC3 with Centos base image.
But in our Dev env we use OC3 with RHEL7 base image. We found there the error.

Issue #999, opened 2021-07-12
Jenkins audit plugin


braisvq1996:
**Is your feature request related to a problem? Please describe.**
Sometimes it’s difficult to tell who is doing what in Jenkins. Something happens, the Jenkins has/need a reset , a pipeline was deleted, and we do not know who did the change.

**Describe the solution you'd like**
Can we consider to install a plugin like this one?
https://plugins.jenkins.io/audit-trail/#documentation

Consider that an audit trail plugin will fill up the disk space much quicker. Jenkins instances are not prepared to store loads of log files so we need to be aware of this in case we install this plugin.

@metmajer @felipecruz91 @jssnull

metmajer:
+1

metmajer:
@braisvq1996 identify an owner, figure out a solution to the disk space problem, and do it, thanks

metmajer:
@braisvq1996 since we're approaching the installation of ODS 4.x, was there any progress?

braisvq1996:
@metmajer , I did not have much time to take a look into it.
We will try to allocate time for this

braisvq1996:
@metmajer we have created a draft pr for this

PR #1043, opened 2021-09-15
Jenkins audit plugin - master

braisvq1996:
Draft for #999


braisvq1996:
The logs generated will look like this

image

braisvq1996:
@metmajer that image was from the first configuration we set, right now it is set to 5 MB *4

This configuration can be changed in a file inside the PVC `${JENKINS_HOME}/audit-trail.xml` or in the UI `${JENKINS_URL}/configure`

Maybe we should set a condition so that the configuration is not overridden when the pod restart [here](https://github.com/opendevstack/ods-core/pull/1043/files#diff-c58b0cedcc974ba11a27f59ea99270487940377706c623f36db7a75ae401fb27R82)

PR #1099, opened 2022-02-10
Modifications needed to be able to upgrade atlassian suite without reinstalling everything.

victorpablosceruelo:
Modifications needed to be able to upgrade atlassian suite without reinstalling everything.

Basically I've included a function "atlassian_stack_reset" that removes the atlassian suite and reinstalls it.

[openshift@ip-10-0-0-50 ods-core]$ ods-devenv/scripts/deploy.sh --branch task/upgrade-atlassian-stack --target atlassian_stack_reset

I've used it to upgrade my box from jira 8.5.8 to jira 8.20.5.

The current license is not valid, but this is a minor issue, since we can enter it manually.


michaelsauter:
Un-assigning myself as I do not know the inner workings of this box, sorry.

PR #1107, opened 2022-02-18
agent base image use ubi repos and cleanup non working

gerardcl:
port of https://github.com/opendevstack/ods-core/pull/1106


tbugfinder:
Currently epel is used in terraform agent. Could this repo be already added in this agent and kept disabled?

gerardcl:
> Currently epel is used in terraform agent. Could this repo be already added in this agent and kept disabled?

why not add this only in terraform if only required there? anyway I will take a look too 👍

gerardcl:
@tbugfinder are you talking about centos7 or ubi8 Dockerfile?

tbugfinder:
> @tbugfinder are you talking about centos7 or ubi8 Dockerfile?

I'd like to focus on ubi8 only.
This can be done in terraform agent itself, just asking for clarification as the base image might be used by different downstream builds.

gerardcl:
then my suggestion would be that meanwhile it is not a general requirement but specific to terraform we add it in the terraform one (I confirm it is only used in this one for now)

PR #1108, opened 2022-02-21
simplify force auth configuration for sonarqube

### LINKED ISSUE #986: Simplify SonarQube config
Instead of forcing auth via API in the configure script (https://github.com/opendevstack/ods-core/blob/master/sonarqube/configure.sh#L199-L205) we can do this in the container startup script: https://github.com/opendevstack/ods-core/blob/master/sonarqube/docker/run.sh.

FYI @felipecruz91

gerardcl:
Closes #986


michaelsauter (dismissed):

gerardcl:
hm @michaelsauter  test is failing...and it is set as expected...will double check

gerardcl:
I think the solution would be to add java arg ` -Dsonar.forceAuthentication=true`
@michaelsauter FYI

gerardcl:
@michaelsauter I have decided to move this fix here https://github.com/opendevstack/ods-core/pull/1096

PR #1121, opened 2022-03-10
fix #1116: add plugins to ubi8 jenkins managed plugins

### LINKED ISSUE #1116: Updating an older Jenkins from 4.7 base image to 4.8/4.9 base image breaks Git plugin
**Describe the bug**
After updating a Jenkins Master from a 4.7 based image to 4.8 based image (UBI8), the Git plugin does not load, which leads to a broken Jenkins instance (no pipeline starts), as discovered with @velcrin @faust2199. 

**Steps To Reproduce**
If the Jenkins Master being upgraded has a [parameterized-trigger-plugin](https://github.com/jenkinsci/parameterized-trigger-plugin) below v0.39, then the update fails. Updating the version of the plugin to v0.39 or higher (tested with v0.43) fixes the problem.

 - OpenShift: 4.8
 - OpenDevStack: 4.x, master

**Additional context**
In https://github.com/opendevstack/ods-core/pull/1040, the version of the plugin is hardcoded in `plugins.txt`, but only for the RHEL7 variant. It may be that for relatively recent Jenkins installs, the version picked for `parameterized-trigger-plugin` is already v0.39+ and therefore updates work.

We suspect the same issue happens for 4.9 based images (and there may be additional errors).

FYI @gerardcl @cschweikert @braisvq1996

velcrin:
Add plugins necessary to upgrade to 4.9 base image in the list of managed plugins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Add support for ods-saas-service quickstarter ([#1033](https://github.com/opendevstack/ods-core/pull/1033))
 - ODS AMI build failing due to broken helm diff package ([#1083](https://github.com/opendevstack/ods-core/pull/1083))
 - ODS AMI CI build fails with Prov APP (ocp check precondition) ([#1117](https://github.com/opendevstack/ods-core/pull/1117))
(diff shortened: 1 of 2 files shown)


michaelsauter (approved):
👍 

@braisvq1996 I think this also makes sense for 4.x, right?

Issue #1125, opened 2022-03-28
Provisioning of projects doesn't setup Jenkins with Aqua credentials 


cschweikert:
### Description:
When provisioning a new project the new Jenkins instance in the new cd project doesn't have credentials of Aqua.

### Expected behavior:
Either the right credentials are set up in Jenkins (`Manage Jenkins > Manage Credentials > ...`) OR no implicit and automatic runs of Aqua in Jenkins pipeline.

### Affected version:
 - OpenShift: 4.8
 - OpenDevStack: 4.x

### Log Output:
```
[...]
[Pipeline] sh (Get JSON of ConfigMap/aqua)
+ oc -n ods get ConfigMap/aqua -o json
[Pipeline] echo
Please provide the alert emails of the Aqua platform!
[Pipeline] echo
Not parameter 'projects.unitt.enabled' at cluster level. Default enabled
[Pipeline] echo
[Aqua Security Scan-fe-angular-test] **** STARTING stage 'Aqua Security Scan' for component 'fe-angular-test' branch 'master' ****
[Pipeline] stage
[Pipeline] { (Aqua Security Scan)
[Pipeline] echo
Please provide the alert emails of the Aqua platform!
[Pipeline] echo
Please provide the name of the repository in Nexus to store the reports!
[Pipeline] echo
Starting to scan via Aqua CLI...
[Pipeline] withCredentials
[Pipeline] // withCredentials
[Pipeline] }
[Pipeline] // stage
[Pipeline] echo
[Aqua Security Scan-fe-angular-test] **** ENDED stage 'Aqua Security Scan' for component 'fe-angular-test' branch 'master' **** (took 113 ms)
[Pipeline] echo
WARN: Error with Aqua due to: org.jenkinsci.plugins.credentialsbinding.impl.CredentialNotFoundException: Could not find credentials entry with ID 'unitt-cd-aqua'
[...]
```

cschweikert:
Another problem with this implicit runs of Aqua are the unit tests of quickstarters. They usually cross-check the names of the build steps within Jenkins pipelines and they don't know about this Aqua step yet.

metmajer:
@braisvq1996 @gerardcl please create the maintenance project finally and add this one

braisvq1996:
> @braisvq1996 @gerardcl please create the maintenance project finally and add this one

As soon as I finish the current task at hand I will start working on the maintenance topic.

[...]

PR #1126, opened 2022-04-07
Point aqua credential id to project-specific cd user

cschweikert:
fixes https://github.com/opendevstack/ods-core/issues/1125

@jorge-romero I don't know what side effects this change might have beside the impact on the issue mentioned above. Maybe you can help with that 🙂


braisvq1996:
From my side, this is the configuration that we use in our ODS installation

metmajer:
@braisvq1996 besides the Changelog is the approach good to go or not?

braisvq1996 (approved):
@metmajer this is ok from my side

PR #1164, opened 2022-07-26
Nexus update to 3.40.1

braisvq1996:
Upgrade of Nexus to version 3.40.1 (current latest)

@michaelsauter @gerardcl


braisvq1996:
Tested in a fresh nexus instance in version 3.40.1 and running the configuration script and another nexus instance in version 3.28.1 that was already configured and making an upgrade.
In both cases there were no issues.

michaelsauter (approved):

Issue #1186, opened 2022-09-28
JAVA_TOOL_OPTIONS cannot be set in docker images. 


victorpablosceruelo:
**Describe the bug**
The line
https://github.com/opendevstack/ods-core/blob/2063d3b4bd46297ac1c46d91bcf66f2a9304b097/jenkins/agent-base/Dockerfile.centos7#L14
might affect java compilations. All java processes will inherit this options, and they are only valid for jenkins agents.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Affected version (please complete the following information):**
 - OpenDevStack [master]

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
N/A

**Additional context**
This was introduced while trying to set default options for java agents.
Need to look for a different way for doing it.

victorpablosceruelo:
@braisvq1996 tomorrow we can continue with the talk about this.

victorpablosceruelo:
The bug was introduced here:
https://github.com/opendevstack/ods-core/pull/1182
https://github.com/opendevstack/ods-core/pull/1182/files#
jenkins/agent-base/Dockerfile.centos7

PR #1190, opened 2022-10-03
Change JAVA_TOOL_OPTIONS by JNLP_JAVA_OPTIONS

victorpablosceruelo:
JAVA_TOOL_OPTIONS is an environment variable that affects all java processes running in the docker image.
JNLP_JAVA_OPTIONS is the environment variable that only affects Jenkins Agents processes.

For obvious reasons, we do not want parameters set for running Jenkins agents to be set for all java processes.

This fixes / closes https://github.com/opendevstack/ods-core/issues/1186


victorpablosceruelo:
Just in case you need to check which file is affected by modification:
https://github.com/openshift/jenkins/blob/master/slave-base/contrib/bin/run-jnlp-client

braisvq1996 (approved):
Good catch !!

Issue #1191, opened 2022-10-27
Update Containers Probes in Jenkins


braisvq1996:
**Is your feature request related to a problem?**
Jenkins takes too long to boot up and if the cluster is under heavy load to the storage provisioner it will take even longer that 15 min.

**Describe the solution you'd like**
In order to avoid Jenkins to restart all the time once this happens we would need to adjust the containers probes and add the [Startup Probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes) 

**Describe alternatives you've considered**
Increase the InitialDelaySeconds parameter

**Additional context**
Currently Jenkins would take 5 min for the container to be in ready state with the current container probes

![image](https://user-images.githubusercontent.com/26645694/198279893-9b954c6d-da9e-4d3d-b2e7-b9fde94c23d3.png)

braisvq1996:
Last test it took 8 min for a jenkins instance to be ready

PR #1195, opened 2023-01-20
Update Containers Probes in Jenkins

braisvq1996:
fix for #1191 

This is needed as we have seen in our biggest cluster issues with Jenkins and its porbes due to the amount of instances

@metmajer @michaelsauter

### DIFF
--- a/jenkins/ocp-config/deploy/jenkins-master.yml
+++ b/jenkins/ocp-config/deploy/jenkins-master.yml
@@ -144,7 +144,7 @@ objects:
     strategy:
       activeDeadlineSeconds: 21600
       recreateParams:
-        timeoutSeconds: 600
+        timeoutSeconds: 1200
       resources: {}
       type: Recreate
     template:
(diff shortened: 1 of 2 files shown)


jenkins/ocp-config/deploy/jenkins-master.yml
```diff
@@ -201,27 +201,34 @@ objects:
...
             httpGet:
               path: /login
               port: 8080
               scheme: HTTP
-            initialDelaySeconds: 3
+            timeoutSeconds: 3
             periodSeconds: 10
             successThreshold: 1
-            timeoutSeconds: 3
+            failureThreshold: 30
```
michaelsauter, line 231: Maybe this `failureThreshold` and the one of the `livenessProbe` could be reduced now that there is a separate `startupProbe`?

michaelsauter (approved):

jenkins/ocp-config/deploy/jenkins-master.yml
braisvq1996: I have made a quick test and we can reduce it safelly to 15 for both probes, but i would maintain the one for the StartupProbe

michaelsauter (approved):

Issue #1224, opened 2023-05-08
Jenkins maintenance


braisvq1996:
**Jenkins maintenance**

- [ ] Update plugins - https://github.com/opendevstack/ods-core/pull/1235
- [ ] Update Jenkins version and aling with base agent - https://github.com/opendevstack/ods-core/pull/1235
- [ ] Align Aquacli version #1219 
- [ ]  Update Git LFS #1206 
- [x]  Remove legacy configuration and plugins - #1225
- [x] Configure durability for max performance - https://github.com/opendevstack/ods-core/pull/1225
- [x] Update resources for deployment (~~Currently testing to ensure estability~~) - https://github.com/opendevstack/ods-core/pull/1234

PR #1234, opened 2023-05-19
Update deployment resources for optimal usage

braisvq1996:
related to #1224

- Reduced Jenkins memory request and limit from 4 Gi to 2560 Mi
- Reduced JVM max metaspace from 1g to 512m
- Reduced JVM heap param from  Xmx2g to Xmx1g

### DIFF
--- a/jenkins/master/Dockerfile.ubi8
+++ b/jenkins/master/Dockerfile.ubi8
@@ -1,4 +1,4 @@
-FROM quay.io/openshift/origin-jenkins:4.11
+FROM quay.io/openshift/origin-jenkins
 
 ENV JAVA_HOME /usr/lib/jvm/jre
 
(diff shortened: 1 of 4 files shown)


braisvq1996:
Jar package seems to have been removed from latest Jenkins image in Quay but not from any Redhat.io registry (openshift4/ose-jenkins or ocp-tools-4/jenkins-rhel8).

jenkins/master/Dockerfile.ubi8
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/openshift/origin-jenkins
```
michaelsauter, line 1: IIRC this has no actual effect due to https://github.com/opendevstack/ods-core/blob/master/jenkins/ocp-config/build/bc.yml#L119.

braisvq1996: True, but for GitHub actions test it does use it right?

PR #1253, opened 2023-12-04
fix allowed change ref types environment read

aoteropas:
change the name of the variable in the if clause for the one that contains the env values. Follow the same approach that other Env Variables


michaelsauter (approved):

braisvq1996:
This change now make Webhook Proxy load the modified parameters but something is not working properly just yet
`````
2023/12/04 10:03:07 zCNVmE Skipping change ref type BRANCH as ALLOWED_CHANGE_REF_TYPES does not include it
2023/12/04 10:03:59 PFYwTB -----
2023/12/04 10:03:59 PFYwTB Skipping change ref type TAG as ALLOWED_CHANGE_REF_TYPES does not include it

jafarre-bi (approved):

Issue #1196, opened 2023-01-20
Remove sonar-crowd plugin in SonarQube


tbugfinder:
The repository for the sonar-crowd plugin was archived.
As there's no development ongoing the plugin should be removed from the tool-stack.

https://github.com/deepy/sonar-crowd/

https://github.com/opendevstack/ods-core/blob/2fcad9cb32d633eb7c50ae1802387b06e10d7338/sonarqube/docker/Dockerfile#L56

braisvq1996:
Yes, it should be removed for the next version of SonarQube (9.x)
One option is to configure SAML with Azure instead of Crowd. LDAP is also an option see documentation [here](https://docs.sonarqube.org/latest/instance-administration/authentication/overview/)

braisvq1996:
SonarQube new LTS version (9.9) have just been released. Once cnes-report plugin gets updated for this version we can also udpate our instances and remove Crowd plugin

Issue #992, opened 2021-06-28
SQ liveness probe missing


michaelsauter:
There is no `livenessProbe` configured for SonarQube, only a readiness probe. The liveness probe should depend on a working database connection, we have seen this to be broken.

Readiness probe is defined here: https://github.com/opendevstack/ods-core/blob/master/sonarqube/ocp-config/sonarqube.yml#L314-L323

gerardcl:
hi @michaelsauter I will take a look into this for ODS 4.2.0

gerardcl:
thanks to @braisvq1996  to provide https://next.sonarqube.com/sonarqube/web_api <- will try to approach the fix via this API

gerardcl:
we will need to add a section in the admin docs that when upgrading sonarqube one needs to disable the probes

braisvq1996:
We can make use of this endpoint https://next.sonarqube.com/sonarqube/web_api/api/system/status as it does not need of any credential

![image](https://user-images.githubusercontent.com/26645694/184144741-d92f5ed5-33ab-439c-ba17-75e3cba2c12a.png)

PR #1267, opened 2024-03-19
SonarQube Maintenance (LTS update, DB update, SAML setup, Helm migration)

### LINKED ISSUE #1211: SonarQube maintenance
Tasks: 
- [x] Change the approach and use the image that vendor provide https://hub.docker.com/_/sonarqube/tags (consume the image the same we use for Nexus)
- [x] Remove Sonar Crowd plugin  and use for example SAML with Azure AD  #1196
- [x] Upgrade SonarQube plugins
- [x] Upgrade SonarQube LTS version (9.9.x)
- [x] Add liveness probe #992 
- [x] Update PostgreSQL version as we currently use 12 and it will reach EOL by the end of 2024 (https://www.postgresql.org/support/versioning/)

**Dependencies**
We have a dependency with an external package (cnes-report) that needs to be updated to support 9.9 version (new version released and supports SQ9.9)

- https://github.com/cnescatlab/sonar-cnes-report/issues/345

braisvq1996:
Closes #1211 #1196 #992 #1268

In this PR we also drop tailor and use helm for Sonarqube

Thanks @ssevillano86 for the initial inplementation


gerardcl:
FYI @henrjk

braisvq1996:
Opened https://github.com/opendevstack/ods-core/issues/1268 as the current backup process is taking over 15 min in a database of 10 GB

docs/modules/administration/pages/helm-migration.adoc
```diff
@@ -0,0 +1,31 @@
+= Migrating from Tailor to Helm
+
+Tailor has been developed for OpenShift 3.11. Back in the days, Helm 2 required the use of a privileged Tiller service and did not work well with OpenShift-specific resources. With Helm 3 and OpenShift 4, this situation as changed.
```
michaelsauter, line 3: s/this situation as changed/this situation has changed

sonarqube/chart/templates/secrets-postgresql.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+data:
+  database-password: {{ .Values.global.sonarDatabasePasswordB64 }}
+kind: Secret
+metadata:
+  annotations:
+    template.openshift.io/expose-database_name: '{.data[''database-name'']}'
+    template.openshift.io/expose-password: '{.data[''database-password'']}'
+    template.openshift.io/expose-username: '{.data[''database-user'']}'
```
michaelsauter, line 9: Are those annotations needed / correct? It looks like the username and DB name are in the configmap?

sonarqube/docker/run.sh
```diff
@@ -2,20 +2,6 @@
...
-    echo "sonar.security.realm=Crowd"
-    echo "crowd.url=$SONARQUBE_CROWD_URL"
-    echo "crowd.application=$SONARQUBE_CROWD_APP"
-    echo "crowd.password=$SONARQUBE_CROWD_PWD"
-    echo "sonar.security.localUsers=admin"
-  } >> conf/sonar.properties
-fi
-
-# enforce authentication to sonar
-echo "sonar.forceAuthentication=true" >> conf/sonar.properties
```
michaelsauter, line 17: Sure this should be removed?

sonarqube/chart/templates/secret.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Secret
+data:
+  admin-password: {{ .Values.global.sonarAdminPasswordB64 }}
+  {{ if ( eq .Values.sonarqube.sonarAuthSaml "true" )}}
+  Saml-Provider-ID: {{ .Values.sonarqube.sonarAuthSamlProviderIdB64 }}
+  Saml-login-url: {{ .Values.sonarqube.sonarAuthSamlLoginUrlB64 }}
+  Saml-cert: {{ .Values.sonarqube.sonarAuthSamlCertficateB64 }}
```
michaelsauter, line 8: Is the casing on purpose?

docs/modules/administration/pages/helm-migration.adoc
```diff
@@ -0,0 +1,31 @@
...
+
+While Tailor also works in an OpenShift 4 cluster, OpenShift has integrated Helm into its product, and Helm has a huge and growing community. Therefore, it is recommended to use Helm instead of Tailor in an OpenShift 4 cluster.
+
+This document will describe how to migrate from Tailor to Helm for OpenDevStack installation.
+
+Tailor is based on OpenShift templates, which define the Kubernetes resources to apply. Helm uses a different templating language / engine, but in the end the templates also describe Kubernetes resources. Therefore, migration effort is relatively low as one only needs to change the syntax of the definition, not the definition itself. Further, there are differences between the CLI of the two tools and not all features of Tailor are available in Helm and vice-versa. Once migration to Helm is complete, it is also recommended to look at the best practices in the Helm community and adopt these.
+
+== Templates preparation
```
michaelsauter, line 11: I was a bit confused by this section initially. At first I thought that `tailor2helm.sh` would convert the OpenShift templates used by Tailor into the chart template files used by Helm. Now I understand this section does not actually provide guidance how to do that conversion, but rather how to deal with the resources already present in the K8s namespace.

michaelsauter (commented):
Nice work 👌

sonarqube/chart/templates/secrets-postgresql.yaml
braisvq1996: True, none of this annotations are needed. Will remove

docs/modules/administration/pages/helm-migration.adoc
braisvq1996: Exactly, when I was preparing to update from a SonarQube installation done with Tailor to a new version using Helm it was not possible to do this except if the old resurces are annotated and labeled for Helm using the script

sonarqube/chart/templates/secret.yaml
braisvq1996: yes, in case you do not want to use SAML and rely on local users you can set the value to false in the confogiration and you will not have any SAML config in your installation

docs/modules/administration/pages/helm-migration.adoc
braisvq1996: Typo fixed

sonarqube/chart/templates/secret.yaml
michaelsauter: Oh I meant upper/lowercase ...

braisvq1996: Ohh, it is on purpose but not because it is required to be in that way so it can be modified as we need/want.
If this is hard to understand or it is better to set the names diferently we can do it

docs/modules/administration/pages/helm-migration.adoc
braisvq1996: I modified the page a bit to avoid this confusion

sonarqube/chart/templates/secret.yaml
braisvq1996: Made the values lowercase

michaelsauter (approved):

gerardcl (approved):
LGTM! 🚀

PR #1288, opened 2024-05-22
Workaround for bitbucket trouble with git-notes

ElfenB:
Context: Bitbucket is generating ghost commits that have the commit message "Notes added by 'git notes add'" if git-notes is used (e.g. with semantic-release). To prevent Jenkins starting builds that will result in failure because there is nothing to build, the build should be stopped when those commits are coming in.

### DIFF
--- a/jenkins/webhook-proxy/main.go
+++ b/jenkins/webhook-proxy/main.go
@@ -262,7 +262,11 @@ func (s *Server) HandleRoot() http.HandlerFunc {
 	type requestBitbucket struct {
 		EventKey   string     `json:"eventKey"`
 		Repository repository `json:"repository"`
-		Changes    []struct {
+		Commit     struct {
+			Id      string `json:"id"`
+			Message string `json:"message"`
+		} `json:"toCommit"`
+		Changes []struct {
 			Type string `json:"type"`
 			Ref  struct {
 				DisplayID string `json:"displayId"`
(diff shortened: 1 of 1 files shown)


braisvq1996:
@ElfenB after a little test this change does not seems to break anything and makes git notes changes not trigger the pipeline

ElfenB:
Perhaps we could achieve a similar result by checking the ref that is provided from Bitbucket?
I am getting builds like this:

image

If the namespace "refs/notes" would be ignored when building that would make more sense than ignoring a commit message? Because notes can't be built really because they are not pointing to some commit with code. So that is an error/bug anyways.
@braisvq1996 What do you think?

ElfenB:
@braisvq1996 As discussed, I adapted the logic it to match on the ref display-id instead of the commit message to be more open for different tools.

braisvq1996:
thanks @ElfenB, I will test it out once I get some time for thsi

oalyman (approved):

metmajer:
@ElfenB, thanks for your contribution and the review, @oalyman. We will schedule this for the next possible release. We will also ensure that there are no side effects by executing our automated tests. You will not need to provide any automated tests for your implementation.

braisvq1996:
@ElfenB @metmajer I managed to get some time to test it on my end and I did not find any issues but after the current release, we can run the whole set of e2e to be sure

metmajer:
@braisvq1996 let's make sure this will be included in the next release possible

metmajer:
@ElfenB please resolve the CHANGELOG.md conflict

braisvq1996 (approved):
tested it and works properly

jafarre-bi (approved):

PR #1309, opened 2024-10-28
Update SonarQube to version 10.7

BraisVQ:
Update SonarQube to 10.7 non LTS version.

It would be better if we can include the upcoming new rust plugin version -> https://github.com/C4tWithShell/community-rust/pull/115

Test:
- [x] CI/CD pipeline scans successfully
- [ ] Ensure cnes report works properly -> Fails with version 5.0.0 of cnes report


BraisVQ:
With the current version of the Cnes report (5.0.0) we are unable to update SonarQube to the version 10.7

PR #1330, opened 2025-03-13
Maintenance March 2025

BraisVQ:
- Update Jenkins and agent to the latest rhel9 version with 4.16 tag. Jenkins plugins and packages also updated.
- Update Webhook proxy to golang 1.24
- Update nexus to the latest 3.70.x version
- Update SonarQube to 2025.1 LTS/LTA version

Potentially we will need to take care of #1331 in this PR

TESTS:

- [x] Docker build and deploy successfully
- [x] Sonarqube/nexus update to new version is correct
- [ ] Component pipeline with Sonarqube and Trivy scan successfully completed
- [ ] Orchestration pipeline with a helm component and a tailor one successfully completed


BraisVQ:
As there is no quay image available for jenkins rhel9/ubi9 we cannot have any test for the build of this image. See:

- https://github.com/openshift/jenkins/issues/1766
- https://github.com/openshift/jenkins/issues/1829

PR #1338, opened 2025-04-02
Added Helm Opentelemetry Collector infrastructure configuration

angelmp01:


BraisVQ (approved):
@angelmp01 check with the latest change I made if it is working properly, but otherwise it is good already

BraisVQ (approved):

PR #1394, opened 2026-07-10
Fix VIT0089535

jafarre-bi:
Validate projectId and projectGroupus to prevent shell metacharacter injection


valituguran (dismissed):
One thing: the changelog comment should say VIT0089535.

valituguran (approved):
All good.

PR #1398, opened 2026-07-13
Fix possible vulnerabilities in webhook proxy

jorge-romero:


jenkins/webhook-proxy/main.go
```diff
@@ -661,11 +661,17 @@ func (c *ocClient) Forward(e *Event, triggerSecret string) (int, []byte, error)
 		e.Pipeline,
 		triggerSecret,
 	)
+	redactedURL := fmt.Sprintf(
+		"%s/namespaces/%s/buildconfigs/%s/webhooks/[REDACTED]/generic",
```
jafarre-bi, line 720: Here it would be more informative to write [SECRET] or [TRIGGER_SECRET]

jenkins/webhook-proxy/main.go
jafarre-bi, line 713: Following the recommendations, we shouldn't embed the trigger secret in the format String

jenkins/webhook-proxy/main.go
```diff
@@ -349,6 +354,11 @@ func (s *Server) HandleRoot() http.HandlerFunc {
 		jenkinsfilePath := jenkinsfilePathDefault
 		jenkinsfilePathParam := queryValues.Get("jenkinsfile_path")
 		if jenkinsfilePathParam != "" {
+			if !safeJenkinsfilePathRegex.MatchString(jenkinsfilePathParam) {
+				log.Println(requestID, "jenkinsfile_path param rejected:", jenkinsfilePathParam)
+				http.Error(w, "Invalid jenkinsfile_path", http.StatusBadRequest)
+				return
+			}
```
jafarre-bi, line 391: This is missing the JSON-escape on template render recommendation

jenkins/webhook-proxy/main.go
```diff
@@ -1091,14 +1091,15 @@ func makePipelineName(project string, component string, branch string) string {
 }
 
 func isProtectedBranch(protectedBranches []string, branch string) bool {
+	lowerBranch := strings.ToLower(branch)
 	for _, b := range protectedBranches {
 		if b == "*" {
 			return true
 		}
-		if strings.HasSuffix(b, "/") && strings.HasPrefix(branch, b) {
+		if strings.HasSuffix(b, "/") && strings.HasPrefix(lowerBranch, b) {
```
jafarre-bi: Should b also be made lowercase?

jenkins/webhook-proxy/main.go
```diff
@@ -344,6 +365,15 @@ func (s *Server) HandleRoot() http.HandlerFunc {
 			return
 		}
 
+		if len(s.AllowedWebhookIPRanges) > 0 {
```
jafarre-bi: Shouldn't we fail if there are no allowed IP ranges?

jafarre-bi (changes requested):
I need to talk to you about a few doubts

jenkins/webhook-proxy/main.go
```diff
@@ -247,6 +256,16 @@ func main() {
 		maxDeletionChecks = envMaxDeletionChecks
 	}
 
+	envAllowedWebhookIPRanges := os.Getenv(allowedWebhookIPRangesEnvVar)
+	if len(envAllowedWebhookIPRanges) == 0 {
+		log.Fatalln(allowedWebhookIPRangesEnvVar, "is required but not set")
+	}
+	allowedWebhookIPRanges, parseErr := parseIPRanges(envAllowedWebhookIPRanges)
+	if parseErr != nil {
```
jorge-romero, line 268: check that it is not empty the list and then it is not needed to check in the lines: 368

jenkins/webhook-proxy/main.go
```diff
@@ -326,18 +349,34 @@ func (s *Server) HandleRoot() http.HandlerFunc {
 		requestID := randStringBytes(6)
 		log.Println(requestID, "-----")
 
+		// TODO(debug): remove before merging
```
jorge-romero: Remove before merge

jenkins/webhook-proxy/main.go
jorge-romero: remove external if as we check it before.

jenkins/webhook-proxy/main.go
jorge-romero: I need to add the equalFold to compare

jenkins/webhook-proxy/main.go
jorge-romero: added lenght check and removed the check later in the code

jenkins/webhook-proxy/main.go
jorge-romero: Removed

jenkins/webhook-proxy/main.go
jorge-romero: The json-escape is done in this way:
a) the template has a filter for the fields "jsonStr" f.e.  "ref": "{{.GitURI | jsonStr}}",
b) the implementation of jsonStr does 2 things, JSON-encodes the string and strips the surrounding quotes so if you provide something like: 
   `main", "admin": true, "evil": "x`

to get a new json like:
```
{
  "ref": "main", 
  "admin": true, 
  "evil": "x"
}
```

you'll get a field value like:
   main\", \"admin\": true, \"evil\": \"x

The final json will be something like:

```
{
  "ref": "main\", \"admin\": true, \"evil\": \"x"
}
```

jenkins/webhook-proxy/main.go
jorge-romero: we cannot have the url in an env variable used by the httpClient object in golang. The url must be provided as it is.
The most it can be done is sanitize the trigger secret.

jenkins/webhook-proxy/main.go
jorge-romero: both lowercase and used equalFold

jafarre-bi (approved):

PR #1399, opened 2026-07-29
Fix jenkins plugins

BraisVQ:
- Fix typo in oidc plugin name that breacks its installation
- Fix git plugin versions, making them compatible with latest git package version


valituguran (commented):
Looks good.

valituguran (approved):

PR #1404, opened 2026-09-01
Clean up configuration

jafarre-bi:


scripts/migrate-provapp-secrets.sh
```diff
@@ -0,0 +1,164 @@
...
+
+        echo "Secret key ${ENV_VAR} already exists, leaving untouched"
+
+    else
+        echo "Adding ${ENV_VAR} to secret"
+
+        oc patch secret "${SECRET_NAME}" \
+          -n "${NAMESPACE}" \
+          --type merge \
+          -p "$(printf '{"stringData":{"%s":"%s"}}' "${ENV_VAR}" "${VALUE}")"
```
jorge-romero: if it si not the value in base64 it can break this. You can use something like:
PATCH=$(jq -n \
  --arg key "$ENV_VAR" \
  --arg value "$VALUE" \
  '{stringData: {($key): $value}}')

oc patch secret "$SECRET_NAME" \
  -n "$NAMESPACE" \
  --type merge \
  -p "$PATCH"

